PR #73, opened 2026-05-03
mmz: fail allocator init when every configured zone is rejected

widgetii:
## Summary

- When the `mmz=` range conflicts with kernel memory (e.g. observed on an `hi3516av200`/V3 camera booted with `mem=256M` while `load_hisilicon` derives `mmz_start=0x88000000,size=384M` from `osmem`/`totalmem` env), `_check_mmz` correctly rejects every zone — but the per-zone loop in `__allocator_init` / `media_mem_parse_cmdline` swallowed the failure and returned 0. `insmod hi_osal.ko` looked successful, the load script's `|| report_error` never fired, every dependent vendor blob loaded on top of an empty MMZ list, and the only symptom was majestic crashing later with no obvious link to the original conflict.
- Track `attempted` vs. `registered` and return `-ENODEV` when the input string was non-empty but no zone survived, so the OSAL/MMZ insmod now fails loudly with the existing `MMZ conflict to kernel memory ...` / `Add MMZ failed: ...` lines still in `dmesg` right above the failure.
- Applied across every allocator that swallowed register failures: V1/V2/V2A legacy MMZ, V3 (`cv300`, `3519v101`) `hisi`+`cma`, V3.5 `cv500` `cma`, and V4 shared pre-5.16 `cma`. V3.5 `hisi`, V4 `allocator.c`, and V4 post-5.16 `cma` already fail fast on register error and were left alone.
- Drive-by in V2/V2A (`cv200`, `av100`): the post-register wraparound check dereferenced `zone` after `hil_mmz_destroy` on the failure path. Restructured to `continue` on register failure (also enables the new counter).
- V1 (`cv100`) `media_mem_init` was discarding the parser return value; capture and propagate it so the new `-ENODEV` actually fails insmod.

## Test plan
- [ ] CI: 8 SDK builds + 8 QEMU smoke tests pass (silent module-load regression already covered by #63).
- [ ] On the affected `hi3516av200`/V3 camera with the conflicting `mem=256M` cmdline + `osmem=128M` U-Boot env: `insmod hi_osal.ko mmz=anonymous,0,0x88000000,384M anony=1` now exits non-zero, `load_hisilicon`'s `|| report_error` fires, and downstream `hi3519v101_*` blobs no longer load on top of an empty MMZ list.
- [ ] Happy-path regression: a correctly-configured camera (e.g. `hi3516ev200` with `mem=32M mmz_allocator=hisi`) still loads `open_osal.ko` and brings up the pipeline.

### DIFF
--- a/kernel/mmz/hi3516av100/media-mem.c
+++ b/kernel/mmz/hi3516av100/media-mem.c
@@ -1026,6 +1026,7 @@ static int media_mem_parse_cmdline(char *s)
 	hil_mmz_t *zone = NULL;
 	char *line;
     unsigned long phys_end = 0;
+	int attempted = 0, registered = 0;
 
 	while( (line = strsep(&s,":")) !=NULL) {
 		int i;
@@ -1063,10 +1064,14 @@ static int media_mem_parse_cmdline(char *s)
 		}
 		mmz_info_phys_start = zone->phys_start + zone->nbytes - 0x2000;
 
+		attempted++;
 		if (hil_mmz_register(zone)) {
 			printk(KERN_WARNING "Add MMZ failed: " HIL_MMZ_FMT_S "\n", hil_mmz_fmt_arg(zone));
 			hil_mmz_destroy(zone);
+			zone = NULL;
+			continue;
 		}
+		registered++;
 
         // if phys_end is 0xFFFFFFFF (32bit)
         phys_end = (zone->phys_start + zone->nbytes);
@@ -1085,6 +1090,12 @@ static int media_mem_parse_cmdline(char *s)
         }
 
 		zone = NULL;
+	}
+
+	if (attempted > 0 && registered == 0) {
+		printk(KERN_ERR "MMZ: all %d configured zone(s) failed to register; refusing to load\n",
+				attempted);
+		return -ENODEV;
 	}
 
 	return 0;
--- a/kernel/mmz/hi3516cv100/media-mem.c
+++ b/kernel/mmz/hi3516cv100/media-mem.c
@@ -957,6 +957,7 @@ static int media_mem_parse_cmdline(char *s)
 {
 	hil_mmz_t *zone = NULL;
 	char *line;
+	int attempted = 0, registered = 0;
 
 	while( (line = strsep(&s,":")) !=NULL) {
 		int i;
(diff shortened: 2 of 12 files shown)


widgetii:
## Live test on `hi3516av200` (V3 family, kernel 3.18.20)

Reproduced the misconfig (kernel `mem=256M` cmdline + load script computing `mmz_start=0x88000000,size=384M` from `osmem=128M`/`totalmem=512M` env), then ran the patched `open_osal.ko` against it.

**First attempt (commit c43f03f only) — only partial fix.** `__allocator_init` correctly returned `-ENODEV` and the new diagnostic appeared in dmesg, but `insmod hi_osal.ko` still succeeded:

```
Hisilicon Media Memory Zone Manager
The module param setup_allocator is hisi
ERROR: Conflict MMZ:
PHYS(0x88000000, 0x9FFFFFFF), GFP=0, nBYTES=393216KB,   NAME="anonymous"
MMZ conflict to kernel memory (0x80000000, 0x8FFFFFFF)
Add MMZ failed: PHYS(0x88000000, 0x9FFFFFFF), ...
MMZ: all 1 configured zone(s) failed to register; refusing to load
```

But `lsmod` showed `open_osal` loaded with vendor blobs piling on top, repeating the original broken state.

Cause: `kernel/osal/{hi3519v101,hi3516cv300,hi3516cv500}/osal.c` was discarding `media_mem_init()`'s return and printing `"hi_osal init success"` regardless. Fixed in 2d81683 — capture the return, unwind `himedia_init`/`osal_proc_init` on failure, propagate up.

**Second attempt (commit 2d81683 added) — fix works end-to-end:**

```
$ insmod hi_osal.ko mmz=anonymous,0,0x88000000,384M anony=1
insmod: can't insert 'hi_osal.ko': No such device
$ echo $?
19    # -ENODEV

$ /usr/bin/load_hisilicon -i
mmz_start: 0x88000000, mmz_size: 384M
hisilicon: Get data from environment and set SENSOR as imx385
++++++++++++++++++++++++++++++++++++++++++++++
==============vi_vpss_online==============
insmod: can't insert 'hi_osal.ko': No such device
******* Error: There's something wrong, please check! *****
$ echo $?
1

$ lsmod
Module                  Size  Used by    Tainted: G
vfat                    8063  0
fat                    50684  1 vfat

$ ps | grep -E 'insmod|rmmod|load_'   # no zombies
```

Compared to **before the patch**, where `lsmod` showed `open_osal` + 8 `hi3519v101_*` blobs partially loaded with `/proc/media-mem` empty, and stuck `insmod hi3519v101_vedu.ko` / `rmmod -w hi3519v101_isp` zombie processes that required `reboot -f` to clear.

Issue #122, opened 2026-05-13
ive_neo (cv500): wire LKOpticalFlowPyr HW dispatch (issue #112 sub-task)


widgetii:
Part of #112 audit (2026-05-13). cv500 vendor blob has full HW dispatch path for LKOpticalFlowPyr but openhisilicon currently silent-stubs the ioctl with `return 0`.

## Vendor symbols (cv500 `obj/hi3516cv500/hi_ive.o`)
- `ive_fill_lk_optical_flow_pyr_task` — task-node field map
- `ive_check_lk_optical_flow_pyr_param` — validator
- `ive_lk_optical_flow_pyr` — top-level handler

## Wire
- ioctl: `0xc2c0461c`
- userspace API: `HI_MPI_IVE_LKOpticalFlowPyr` — already exported.

## Description
Lucas-Kanade pyramid optical flow tracking. Takes prev/next image pyramid arrays + prev point list, outputs next-point positions, per-point status, per-point error. cv500 vendor sample `sample_ive_st_and_lk.c` exercises this end-to-end so the API + ctrl shape are well-documented.

## Definition of done
- [ ] Field map decoded, captured to kaeru as `ive-neo-cv500-lk-field-map`.
- [ ] Kernel handler `ive_op_lk_optical_flow_pyr_cv500()` wired; silent-stub case replaced.
- [ ] Userspace marshalling — multi-level pyramid array packing into kernel arg buffer.
- [ ] Round-trip test (single frame static → all points report 'tracked OK' at same coords).
- [ ] CI needle-grep extended.

## Related
Vendor sample at `Hi3516CV500_SDK_V2.0.2.1/smp/a7_linux/mpp/sample/svp/ive/sample/sample_ive_st_and_lk.c` — useful reference for the userland flow + control struct values.

Issue #121, opened 2026-05-13
ive_neo (cv500): wire Resize HW dispatch (issue #112 sub-task)


widgetii:
Part of #112 audit (2026-05-13). cv500 vendor blob has full HW dispatch path for Resize but openhisilicon currently silent-stubs the ioctl with `return 0`.

## Vendor symbols (cv500 `obj/hi3516cv500/hi_ive.o`)
- `ive_fill_resize_task` @0x91e8 — task-node field map (multi-N loop over src/dst arrays)
- `ive_check_resize_param` @0x13494 — validator
- `ive_resize_proc` @0x490c — top-level handler

## Wire
- ioctl: `0xc008462e`
- userspace API: `HI_MPI_IVE_Resize(handle, astSrc[], astDst[], ctrl, instant)` — already exported.

## Description
Multi-N image resize. ctrl is `IVE_RESIZE_CTRL_S` = `{ enMode, stMem (IVE_MEM_INFO_S), u16Num }`. Supports U8C1 and U8C3_PLANAR input/output, bilinear interpolation, 16-pixel-aligned strides. Image size range 32x16 to 1920x1080 per SDK header.

## Description
This is the second-largest op family to wire after KCF (multi-N like PerspTrans/Hog, plus an intermediate `stMem` buffer the HW needs). Similar effort to the PerspTrans HW dispatch landed in #107.

## Definition of done
- [ ] Field map decoded (multi-N loop, intermediate stMem semantics), captured to kaeru as `ive-neo-cv500-resize-field-map`.
- [ ] Kernel handler `ive_op_resize_cv500()` wired; silent-stub case replaced.
- [ ] Userspace marshalling (\`libraries/ive_neo/src/ive_ops.c\`): pack \`astSrc[1..N]\` + \`astDst[1..N]\` + ctrl into the kernel arg buffer.
- [ ] Round-trip test (identity-resize: same dims → dst ≈ src).
- [ ] CI needle-grep extended.

Issue #120, opened 2026-05-13
ive_neo (cv500): wire FilterAndCSC HW dispatch (issue #112 sub-task)


widgetii:
Part of #112 audit (2026-05-13). cv500 vendor blob has full HW dispatch path for FilterAndCSC but openhisilicon currently silent-stubs the ioctl with `return 0`.

## Vendor symbols (cv500 `obj/hi3516cv500/hi_ive.o`)
- `ive_fill_filter_and_csc_task` — task-node field map
- `ive_check_filter_and_csc_param` — validator
- `ive_flt_csc` — top-level handler (seen in vendor dispatcher at `0x6524`)

## Wire
- ioctl: `0xc0c04603`
- userspace API: `HI_MPI_IVE_FilterAndCSC` — already exported.

## Description
5x5 filter + color-space conversion combined op. Single src + single dst + ctrl with 25-byte mask + CSC mode byte.

## Definition of done
- [ ] Field map decoded, captured to kaeru.
- [ ] Kernel handler wired; silent-stub case replaced.
- [ ] Round-trip test (identity filter → dst ≈ CSC(src)).
- [ ] CI needle-grep extended.

Issue #119, opened 2026-05-13
ive_neo (cv500): wire CSC HW dispatch (issue #112 sub-task)


widgetii:
Part of #112 audit (2026-05-13). cv500 vendor blob has full HW dispatch path for CSC but openhisilicon currently silent-stubs the ioctl with `return 0`.

## Vendor symbols (cv500 `obj/hi3516cv500/hi_ive.o`)
- `ive_fill_csc_task` — task-node field map
- `ive_check_csc_param` — validator
- `ive_csc` — top-level handler

## Wire
- ioctl: `0xc0a04602`
- userspace API: `HI_MPI_IVE_CSC` — already exported from `libive_neo.so`, currently calls into a kernel path that returns 0 silently.

## Description
Color-space conversion (YUV↔RGB family). Single src + single dst + ctrl carrying CSC mode. The prior "VGS-only" comment in the dispatcher was wrong for cv500 — vendor blob has a dedicated IVE HW path.

## Definition of done
- [ ] Field map decoded from vendor blob, captured to kaeru as `ive-neo-cv500-csc-field-map`.
- [ ] Kernel handler `ive_op_csc_cv500()` wired; silent-stub case replaced.
- [ ] On-target round-trip test in `libraries/ive_neo/test/test_persp_hog.c` — minimum: HW wrote dst, board stays up.
- [ ] CI: extend `Verify ive_neo built for this platform` needle-grep with the new handler's one-time log string.

## Related
PR #107 (PerspTrans + Hog), PR #115 (NCC), PR #117 (LBP) for the reference pattern.

PR #123, opened 2026-05-13
kernel/ive_neo: audit silent-stub dispatcher cases against cv500 vendor blob

widgetii:
Refresh the silent-stub comments in \`ive_neo.c\`'s dispatcher with evidence-backed findings from cross-referencing the cv500 vendor blob \`obj/hi3516cv500/hi_ive.o\` symbol table against each \`return 0;\` case. **No behaviour change** — comment-only refactor that closes the audit gap from #112.

### (a) cv500 HW supports it, just unwired in openhisilicon
Vendor blob has full \`ive_check_<op>_param\` + \`ive_fill_<op>_task\`. The prior \"VGS-only\" comment for CSC/FilterAndCSC/Resize was wrong for cv500 — they have dedicated IVE HW paths.

Follow-up issues opened:
- #119 — CSC (\`ive_fill_csc_task\` + \`ive_csc\`)
- #120 — FilterAndCSC (\`ive_fill_filter_and_csc_task\` + \`ive_flt_csc\`)
- #121 — Resize (\`ive_fill_resize_task\` + \`ive_resize_proc\` — multi-N)
- #122 — LKOpticalFlowPyr (\`ive_fill_lk_optical_flow_pyr_task\` + \`ive_lk_optical_flow_pyr\`)

### (b) cv500 reuses another op's HW path with a flag override
EqualizeHist calls \`ive_fill_hist_task\` with \`node[7]=0x61\` + an extra LUT remap at \`node[200]\` (decoded from \`ive_ioctl\` dispatcher branch \`0x68d0..0x69d0\`). Distinct from (a) because the handler has no own \`fill_task\` symbol.

- EqualizeHist — #112 umbrella follow-up.
- GMM (single-Gaussian) — GMM2 with default params; vendor only ships \`fill_gmm2_task\`.

### (c) cv500 vendor blob has no symbols whatsoever
HW unit either doesn't exist or vendor never wired userspace to it. Silent-stub \`return 0\` is correct here — libive presumably falls back to CPU or returns its own error.

- GradFg, MatchBgModel, UpdateBgModel
- ANN_MLP_Predict, SVM_Predict

## Method
\`nm obj/hi3516cv500/hi_ive.o | grep -iE \"ive_(check_<op>|fill_<op>|<op>_proc)\"\` for each of the 13 silent-stub ops, cross-checked against the SDK header signatures and (where available) the vendor sample code at \`Hi3516CV500_SDK_V2.0.2.1/smp/a7_linux/mpp/sample/svp/ive/sample/\`.

## What's NOT in this PR
No new handlers, no userspace changes, no test changes. Pure documentation + four new follow-up issues to track each \"(a)\" op.

### DIFF
--- a/kernel/ive_neo/ive_neo.c
+++ b/kernel/ive_neo/ive_neo.c
@@ -2590,20 +2590,49 @@ static long ive_dispatch(unsigned int cmd, unsigned long arg)
 	case 0xc0c0462bu:      return ive_op_st_candi_corner(arg); /* STCandiCorner */
 	case 0xc0f04619u:      return ive_op_canny_hys(arg);  /* CannyHysEdge */
 	case 0xc1a8462du:      return ive_op_gmm2(arg);       /* GMM2 (best effort) */
-	/* Not supported on Hi3516EV200/EV300 — vendor kernel also rejects
-	 * these with "ive can't support the func". Return 0 to keep
-	 * libive.so's handle check happy; output will be empty. */
 	case 0xc0a8461au:      return ive_op_lbp(arg);   /* LBP — vendor op 26 */
 	case 0xc0b84616u:      return ive_op_ncc(arg);   /* NCC — vendor op 22 */
-	case 0xc0b8462au:      return 0;  /* EqualizeHist */
-	case 0xc0a04602u:      return 0;  /* CSC (VGS) */
-	case 0xc0c04603u:      return 0;  /* FilterAndCSC (VGS) */
-	case 0xc008462eu:      return 0;  /* Resize (VGS) */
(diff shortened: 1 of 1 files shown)


widgetii:
Superseded by the actual handler implementations that landed after this audit was drafted:

- (a) CSC / FilterAndCSC / Resize / LK — wired with real HW dispatch in #124, #126, #127.
- (b) EqualizeHist — wired in #128.
- (c) GMM (single-Gaussian) — promoted from (b) to (c) in #128 after grep-confirming cv500 vendor `ive_ioctl` has no case for ioctl-nr `0x4618`.

The three-tier classification framework that this PR introduced still describes the dispatcher state, but the comments live in the merged `ive_neo.c` rather than in this diff.

Resolution tracked at #112.

PR #150, opened 2026-05-17
kernel/hi3516cv300: fix cv300_lite (kernel 3.18) build regression from #148

widgetii:
## Summary

#148 brought cv300 OSAL/peripherals up to Linux 7.0 and was verified end-to-end on kernel 7.0 under QEMU — but it broke the cv300_lite 3.18 path that ships in production firmware. Two distinct root causes, both surfaced while building \`hi3516cv300_lite\` from OpenIPC/firmware#2103:

1. \`mmz/media-mem.c\` called \`p4d_offset()\` unconditionally. The symbol was added in 4.11; on 3.18 it doesn't exist. Gate it with \`LINUX_VERSION_CODE >= 4.11\` and fall back to \`pud_offset(pgd, virt)\` on older kernels — same pattern \`hi3516cv500/mmz/media_mem.c\` and \`osal/linux/kernel/mmz/media-mem.c\` already use.

2. \`wdt/hi_wdt.c\`, \`ir/hiir.c\`, \`pwm/pwm.c\`, \`rtc/hi_rtc.c\` got \`MODULE_LICENSE("GPL")\` lines in #148 without an explicit \`#include <linux/module.h>\`. On 7.0 the include chain via \`hi_osal.h\` transitively pulls module.h; on 3.18 it doesn't, the macro is undefined, and the C preprocessor leaves a bare string-literal call. Add the explicit include — matching what the cv300 init/*.c files that already have \`MODULE_LICENSE\` do.

## Test plan

Built \`hi3516cv300_lite\` from OpenIPC/firmware feat/cv300-neo with this fix pinned in, diffed against today's nightly:

| | with fix | nightly |
|---|---|---|
| rootfs file count | 674 | 674 |
| rootfs file list diff | — | empty |
| hisilicon \`.ko\` set | identical | identical |
| QEMU -M hi3516cv300 lsmod | 5 modules | 5 modules |
| lsmod content diff | — | identical |
| dmesg diff (scrubbed) | — | 4 lines: build host, build time, +/-4K memory rounding, BogoMIPS jitter, random MAC |
| Oops/panic/BUG/Trace | none | none |

No behavioural change to the 7.0 (neo) path either — the fix is conditional (\`#if LINUX_VERSION_CODE\` for p4d; bare include addition for module.h). Both kernel versions take the same compiled paths after this change.

### DIFF
--- a/kernel/compat/kernel_compat.h
+++ b/kernel/compat/kernel_compat.h
@@ -177,13 +177,18 @@ static inline void compat_do_gettimeofday(struct compat_timeval *tv)
 #endif
 
 /*
- * print_symbol() removed in 4.x — was a wrapper around printk("%pS").
- * Define as a transparent macro so legacy OSAL osal_addr.c calls keep
- * working unchanged. fmt is expected to contain "%s" (the old contract);
- * we redirect that single conversion to %pS (kernel symbol lookup).
- * Multi-arg fmt strings are not supported — none of the call sites use them.
+ * print_symbol() removed in 5.15 (commit f74cd6432cb6) — was a wrapper
+ * around printk("%pS"). Define as a transparent macro so legacy OSAL
+ * osal_addr.c calls keep working unchanged. fmt is expected to contain
+ * "%s" (the old contract); we redirect that single conversion to %pS
+ * (kernel symbol lookup). Multi-arg fmt strings are not supported —
+ * none of the call sites use them. Guard at 5.15.0, NOT 4.0.0: the
+ * symbol is still declared in <linux/kallsyms.h> on 4.x and 5.0-5.14,
+ * so redefining as a do-while macro collides with the function prototype
+ * (cv500/cv200 lite builds against 4.9 fail with "expected identifier
+ * or '(' before 'do'").
  */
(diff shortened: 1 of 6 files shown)


widgetii:
Additional fixes pushed: PR #148 also broke cv500_lite / cv200_lite (kernel 4.9, \`print_symbol\` collision) and ev300_neo / av300_neo (kernel 6.4+, \`DEFINE_SEMAPHORE\` collision). Surfaced by the CI matrix run on this branch.

- \`print_symbol\` shim gated \`>= 4.0\` but the kernel function is declared in \`<linux/kallsyms.h>\` through 5.14. Bumped gate to \`>= 5.15\` (commit f74cd6432cb6 removed it).
- \`DEFINE_SEMAPHORE\` shim \`#undef\`+redefined the kernel macro as 1-arg on \`>= 6.4\`, breaking cv500/osal-linux \`compat_DEFINE_SEMAPHORE\` callers which expanded to the 2-arg form. Dropped the \`#undef\`; cv300 osal updated to use \`compat_DEFINE_SEMAPHORE\` like cv500 mmz already does.

Locally verified: cv300_lite still byte-equivalent to nightly (rootfs file list 674 = 674, hisilicon .ko set unchanged) and the other shim-affected platforms should now compile.

Issue #26, opened 2025-05-17
mis2008 sensor lib


vobukay:
**It seems values that calculated by sensor_mirror_flip are incorrect.**

<details><summary>Details</summary>
<p>

[mis2008 datasheet](http://www.szlvdu.com/uploads/202107/60f7efbc821fa.pdf)

```c

static GK_VOID sensor_mirror_flip(VI_PIPE ViPipe, ISP_SNS_MIRRORFLIP_TYPE_E eSnsMirrorFlip)
{
    ISP_SNS_STATE_S *pstSnsState = GK_NULL;
    MIS2008_SENSOR_GET_CTX(ViPipe, pstSnsState);
    CMOS_CHECK_POINTER_VOID(pstSnsState);
    GK_U8 FRAME_H_ST, FRAME_H_END;
    GK_U8 FRAME_W_ST, FRAME_W_END;
    GK_U8 value = 0;
    switch (eSnsMirrorFlip) {
    case ISP_SNS_NORMAL:
        value |= 0;
        FRAME_H_ST = 4;
        FRAME_H_END = 0x3f;
        FRAME_W_ST = 8;
        FRAME_W_END = 0x88;
        break;
    case ISP_SNS_MIRROR:
        value |= 0x01;
        FRAME_H_ST = 4;
        FRAME_H_END = 0x3f;
        FRAME_W_ST = 0xa;
        FRAME_W_END = 0x89;
        break;
    case ISP_SNS_FLIP:
        value |= 0x02;
        FRAME_H_ST = 5;
        FRAME_H_END = 0x40;
        FRAME_W_ST = 8;
        FRAME_W_END = 0x88;
        break;
    case ISP_SNS_MIRROR_FLIP:
        value |= 0x03;
        FRAME_H_ST = 5;
        FRAME_H_END = 0x40;
        FRAME_W_ST = 0xa;
        FRAME_W_END = 0x89;
        break;
    default:
        return;
    }

    mis2008_write_register(ViPipe, 0x3205, FRAME_H_ST);
    mis2008_write_register(ViPipe, 0x3207, FRAME_H_END);
    mis2008_write_register(ViPipe, 0x3209, FRAME_W_ST);
    mis2008_write_register(ViPipe, 0x320b, FRAME_W_END);
    mis2008_write_register(ViPipe, MIS2008_FLIP_MIRROR_ADDR, value);
}
``` 

translated from chinese

[...]

PR #153, opened 2026-05-17
libraries/nnie_neo: marshal ForwardWithBbox to /dev/nnie

widgetii:
Stacked on top of #152 (kernel-side bbox stub).

## Summary
- Replace the userspace-only stub for `HI_MPI_SVP_NNIE_ForwardWithBbox` with the real 1728 B arg-buffer marshalling, so the lib now talks to the kernel ioctl exactly like vendor libnnie.so.
- Public API behaviour is unchanged today (kernel returns -EOPNOTSUPP → translated to `HI_ERR_SVP_NNIE_NOT_SURPPORT`), but the wrapper is now correctly hooked up for when the kernel-side bbox path lands.

## Why this matters now
Before: an app calling `HI_MPI_SVP_NNIE_ForwardWithBbox` against libnnie_neo never reached the kernel — the lib returned NOT_SURPPORT in userspace. That meant kernel instrumentation, ftrace, or future bbox impl work couldn't be exercised through our lib. Behaviour also silently diverged from vendor libnnie at the syscall boundary.

After: the wrapper marshals all six args per the SDK-declared `SVP_NNIE_FORWARD_WITHBBOX_CTRL_S` layout and ioctl's. Kernel-side rejection produces the same `HI_ERR_SVP_NNIE_NOT_SURPPORT` it did before, via the existing `nnie_err_to_hi` translator.

## Test plan
- [x] Cross-build: `make CC=arm-openipc-linux-gnueabi-gcc` cleanly produces libnnie_neo.so/.a.
- [x] CI: new build step in `libraries-cross-compile` compiles `test_forward_with_bbox` and asserts the bbox symbol is exported.
- [x] On-target (av300): with `open_nnie_neo.ko` loaded, `./test_forward_with_bbox` reports PASS; dmesg shows the kernel-side `pr_info_once`:
  ```
  PASS: bbox wrapper marshalled cleanly, kernel returned NOT_SURPPORT
  dmesg: nnie_neo: ForwardWithBbox not implemented yet; use Forward for classifier nets
  ```
- [x] Regression: mnist (`sample_nnie_main 4`) and SSD (`sample_nnie_main 5`) still byte-identical to vendor on av300.

## Out of scope
- Kernel-side bbox handler (#26) — the tskbuf §3 proposal-phys table layout still needs RE; a speculative impl was tested in this branch and hung the HW, requires a board reboot.
- libnnie_neo `HI_MPI_SVP_NNIE_GetTskBufSize` proper implementation — still a stub; would need a per-segment instruction-stream walker.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -570,6 +570,19 @@ jobs:
               fi
           done
 
+      - name: Cross-compile NNIE bbox marshalling test
+        # Target-only smoke test (needs /dev/nnie) — we just verify it
+        # links against libnnie_neo so a missing/renamed symbol in the
+        # bbox wrapper trips CI rather than waiting for on-target.
+        run: |
+          make -C libraries/nnie_neo/test test_forward_with_bbox \
+              CC=arm-linux-gnueabi-gcc
+          file libraries/nnie_neo/test/test_forward_with_bbox | grep -q ARM
+          arm-linux-gnueabi-nm libraries/nnie_neo/test/test_forward_with_bbox \
+              | grep -qE ' T HI_MPI_SVP_NNIE_ForwardWithBbox$' \
+              || { echo "FAIL: bbox symbol missing from test binary" >&2; exit 1; }
+          echo "ok: bbox test links cleanly"
+
   #
   # qemu-ive-ops
   #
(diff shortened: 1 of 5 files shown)


widgetii:
Closing — this PR is behaviour-neutral. The kernel ioctl returns -EOPNOTSUPP either way, so the userspace HI_MPI_SVP_NNIE_ForwardWithBbox caller sees the same HI_ERR_SVP_NNIE_NOT_SURPPORT before and after this change. The marshalling work should be bundled with the kernel-side bbox implementation when that actually lands, not run through CI on its own. Shouldn't have opened it independently.

Issue #58, opened 2026-04-19
ISP source compilation: umbrella tracker for all platforms


widgetii:
## Overview

Tracking ISP subsystem openness across all HiSilicon platforms. The ISP pipeline spans kernel modules (`open_isp.ko`, `open_sensor_i2c.ko`, `open_mipi_rx.ko`) and userspace libraries (`libisp.so`, `libsns_*.so`). The goal is to compile as much as possible from source instead of shipping vendor binary blobs.

The key architectural split is between **V1–V3A** (ISP algorithms in the kernel module) and **V4** (algorithms in userspace `libisp.so`). Source code exists for most platforms but is not always wired into the build.

## Platform Comparison

| Platform | Gen | ISP kernel `.ko` | ISP userspace `libisp.so` | Sensors `libsns_*.so` | MIPI RX | VI | sensor_i2c |
|----------|-----|------------------|--------------------------|----------------------|---------|-----|------------|
| **hi3516ev200** | V4 | **Source** (HAL+shared mkp) | **Source** (41 .c) | 30 sensors, source | Source | Blob | Source |
| **gk7205v200** | V4 | symlink→ev200 | symlink→ev200 | symlink→ev200 | shared | Blob | shared |
| **hi3516cv500** | V3.5 | **Source** (7 .c) | **Source** (41 .c) | 20 sensors, source | Source | Blob | Source |
| **hi3516cv200** | V2 | **Source** (6 .c) | Not built (34 .c exist) | 10 sensors, source | Source | Blob | Source |
| **hi3516av100** | V2A | Blob (33 .c exist) | Not built (25 .c exist) | 14 sensors, source | Source | Blob | Source |
| **hi3516cv300** | V3 | Blob (43 .c exist) | Not built (38 .c exist) | 9 sensors, source | Source | Blob | Source |
| **hi3519v101** | V3A | Blob (40 .c exist) | Not built (35 .c exist) | 9 sensors, source | Source | Blob | Source |
| **hi3516cv100** | V1 | Blob (no source) | None (no source) | 16 sensors, source | None | Blob | Blob |

## Where 3A Algorithms Live

The ISP algorithms (AE, AWB, AF, demosaic, sharpening, DRC, noise reduction, etc.) are the core image quality code. They live in different places depending on chip generation:

| Generation | Platforms | Algorithms location | Count |
|-----------|-----------|-------------------|-------|
[...]

Issue #82, opened 2026-05-05
Sensor source compilation: umbrella tracker for all platforms


widgetii:
## Overview

Tracking sensor-driver (`libsns_*.so`) openness across all HiSilicon / Goke platforms supported by this repo. The goal is to compile every sensor we have source code for from source, instead of shipping the vendor binary blobs that currently live in OpenIPC/firmware under `general/package/{hisilicon,goke}-osdrv-<chiparch>/files/sensor/`. Companion to #58 (ISP).

The pattern was established in [OpenIPC/firmware#2042](https://github.com/OpenIPC/firmware/pull/2042) for `hi3516cv200`: source-built sensors are installed by the `hisilicon-opensdk` package from `HISILICON_OPENSDK_SENSORS_<chiparch>`; the matching binary blobs are deleted from the `osdrv` package; sensors without source stay as binary blobs (no regression). End-to-end verified on a real hi3518ev200 camera — RTSP stream up, `/image.jpg` returns valid 1920×1080 JPEG.

## Per-platform state

| Platform | Gen | Source dirs | Wired in firmware | Gap | Notes |
|---|---|---|---|---|---|
| hi3516cv200 | V2 | 18 | 18 | ✅ DONE | landed in [firmware#2042](https://github.com/OpenIPC/firmware/pull/2042) |
| hi3519v101 | V3A | 9 | 1 | +8 | install foreach already wired |
| hi3516cv500 | V3.5 | 20 | 3 | +17 | install foreach already wired |
| hi3516ev200 | V4 | 30 | 3 | +27 | uses generic `else` install block |
| gk7205v200 | V4-GK | 30 (→ev200) | 3 | +27 | shares ev200 source via symlink |
| hi3516cv300 | V3 | 9 | 0 | +9 | install block needs `foreach` added |
| hi3516av100 | V2A | 14 | 0 | +14 | install block needs `foreach` added |
| hi3516cv100 | V1 (T1+) | 17 | 0 | +17 | install block needs `foreach` added; T1+ flavour landed in #77 |

Numbers correspond to: `ls -d libraries/sensor/<chiparch>/*/ | wc -l` (source dirs) vs. count of `libsns_*` entries in `HISILICON_OPENSDK_SENSORS_<chiparch>` in the firmware `hisilicon-opensdk.mk` (wired).

### Per-platform binary deletion counts

[...]

PR #155, opened 2026-05-17
Add openipc_frame_ts: sensor frame-start (PTS, CLOCK_REALTIME) chrdev

### LINKED ISSUE #144: vi/mipi_rx: capture (encoder PTS, CLOCK_REALTIME) pairs at sensor frame-start IRQ
### Problem

For consumers that need to anchor an encoded frame to wall-clock time (RTSP/RTCP Sender Report NTP timestamps, multi-sensor fusion in ROS 2, evidence chains, OSD timecode), the only timestamp surfaced today is the encoder PTS returned in `VENC_STREAM_S.u64PTS`, which is monotonic since encoder init and not anchored to `CLOCK_REALTIME`. There is no kernel-side capture of "what wall-clock value was current at the moment of sensor frame-start for frame N".

A userspace daemon (e.g. `majestic` or a sidecar) can sample `CLOCK_REALTIME` when it later receives the frame from `HI_MPI_VENC_GetStream`, but by then the value carries the full pipeline delay (ISP processing + encoder buffering + scheduler latency), with tens of milliseconds of variable bias. The 3-5 ms wall-clock accuracy the camera's `ntpd` already maintains is wasted because the per-frame stamp arrives that much later than the actual capture.

### Background and prior art in this tree

The sensor-side frame-start IRQ is observable at two well-defined points:

- **MIPI RX** — `kernel/mipi_rx/mipi_rx_hal.c` registers a shared IRQ via `osal_request_irq(mipi_rx_irq_num, mipi_rx_interrupt_route, ...)`. Per-SoC variants (e.g. `kernel/mipi_rx/hi3519v101/mipi.c`) register `mipi0_int`/`mipi1_int` and dispatch through `mipi_interrupt_route`. This is the **earliest** point in the pipeline where the kernel sees a frame begin — physically the closest moment to sensor exposure end.
- **ISP** — `kernel/isp/arch/hi3519v101/firmware/drv/isp.c` registers `ISP_ISR` as the top-half and `ISP_IntUserBottomHalf` as a thread function. Triggers slightly after MIPI RX (post-ISP-statistics-collection), but the existing top-half / bottom-half split is **the right architectural pattern** for our use case — minimal IRQ-context work, deferred formatting and wakeups.

Note: \`kernel/hifb/hi3516cv500/hifb_vsync.c\` is the *display* VSYNC chrdev — panel scan-out, not sensor capture — and is **not** prior art for this work.

Per-SoC subdir convention in \`kernel/vi/\` groups EV200/EV300 under \`hi3516ev200\`, plus separate \`gk7205v200\`, \`hi3519v101\`, and the \`cv200/cv300/cv500\` family.

A search of the tree shows no current use of \`ktime_get_real_ns()\` or \`ktime_get_clocktai()\` — there is no existing kernel-side wall-clock capture path to extend.

### Proposed approach

Add a small kernel module (or extend the MIPI RX / ISP drivers) that:

1. **Hooks the sensor frame-start IRQ.** Preferred attach point is the existing MIPI RX \`mipi_rx_interrupt_route\` (earliest, closest to true exposure). Fallback / alternative is the ISP \`ISP_ISR\` top-half — slightly later but reuses the existing \`ISP_IntUserBottomHalf\` workqueue, which is the right pattern.
2. **In IRQ context, capture two values:**
   - The encoder PTS source (same \`HI_U64\`-microsecond counter that \`VENC_STREAM_S.u64PTS\` is derived from — typically \`MMP_TIMER\` or equivalent free-running counter).
   - \`ktime_get_real_ns()\` (NTP-disciplined wall clock).
   
   Push the \`(pts_us, wall_ns, vi_chn, seq)\` tuple into a small lock-free ring buffer (default depth 64 entries per channel). Keep IRQ-context work to a register read + ring push only.
3. **Wake userspace from a bottom half / workqueue** following the \`ISP_IntUserBottomHalf\` pattern.
4. **Expose the ring to userspace** via a chrdev (naming TBD — \`/dev/openipc-frame-ts\` or \`/dev/openipc/vi-fs\`). \`read()\` returns packed \`struct openipc_frame_ts_event { __u64 pts_us; __u64 wall_ns; __u32 vi_chn; __u32 seq; }\`. \`poll()\`/\`select()\` supported.
5. **Per-SoC support.** Ship initially for SoCs whose VI / MIPI RX trees already live in the repo: \`hi3516ev200\` (covers EV200/EV300), \`hi3516cv200\`, \`hi3516cv300\`, \`hi3516cv500\`, \`hi3519v101\`, \`gk7205v200\`. The per-SoC differences are confined to *where* \`mipi_rx_irq_num\` is sourced and the channel-count constants; the chrdev surface and ring-buffer logic are shared. Gate behind a kbuild flag \`CONFIG_OPENIPC_FRAME_TIMESTAMP=m\` so consumers can disable.
6. **Document** the new chrdev in a short README under \`kernel/openipc_frame_ts/\` and reference it from the top-level repo README.
7. **Provide a userspace test program** under \`samples/openipc_frame_ts/\` that opens the chrdev, prints incoming pairs at the configured fps, and produces a jitter histogram. Doubles as regression test and worked example.

### Constraints / non-goals

- Must not modify the binary MPI ABI used by \`majestic\` or other Prosperity-PL binaries. The new chrdev is purely additive.
- Must not regress existing \`u64PTS\` values returned by \`HI_MPI_VENC_GetStream\`. This work *complements*, not replaces, the existing PTS path.
- Wall-clock anchor uses \`CLOCK_REALTIME\` (NTP-disciplined by the existing BusyBox \`ntpd\`). PTP would substitute \`ktime_get_clocktai()\` later if needed; not in scope.
- IRQ-context work must remain trivial (counter read + ring push). All formatting/poll-waking moves to a tasklet or workqueue, following the \`ISP_IntUserBottomHalf\` pattern.

### Definition of done

1. On a Hi3516EV300 with a sensor running at 30 fps, a userspace process reading the chrdev receives one event per frame, with \`wall_ns\` deltas matching 33.33 ms ±100 µs over a 60-second window (measured σ < 100 µs).
2. The reported \`pts_us\` matches the value subsequently surfaced in that frame's \`VENC_STREAM_S.u64PTS\`. Verified with a small consumer that subscribes to both frame-start events and VENC output and correlates by PTS.
3. The userspace test program at \`samples/openipc_frame_ts/\` runs without errors and produces a jitter histogram on stdout.
4. Building OpenIPC firmware with this module enabled does **not** break majestic's existing RTSP streams (regression-tested visually and via packet capture).
5. Same as (1)–(4) on at least one additional SoC (Hi3516CV500 preferred — fully shared MIPI RX HAL).
6. Short documentation in the repo README explaining how to enable the module, open the chrdev, and interpret events.

### Out of scope

- Patching \`smolrtsp\` / \`majestic\` to consume this for NTP-anchored RTCP Sender Reports — belongs in [OpenIPC/smolrtsp](https://github.com/OpenIPC/smolrtsp) and [OpenIPC/majestic](https://github.com/OpenIPC/majestic).
- A user-friendly libopenipc API wrapping the chrdev — could happen later in [OpenIPC/firmware](https://github.com/OpenIPC/firmware).
- gPTP / PTP hardware timestamping is a separate question if Hi3516 MAC PTP support is ever explored.

widgetii:
Closes #144.

## Summary

- Adds `kernel/openipc_frame_ts/` — shared chrdev module exposing per-channel ring of `(pts_us, wall_ns)` tuples at `/dev/openipc-frame-ts`. Exports `openipc_frame_ts_push(vi_chn)` for IRQ-context callers.
- Adds `include/openipc_frame_ts.h` — stable userspace ABI (`struct openipc_frame_ts_event` + 2 ioctls).
- Adds per-SoC IRQ hooks that call `openipc_frame_ts_push` at sensor frame-start.
- Adds `samples/openipc_frame_ts/` — userspace test with per-second summary mode.

## Why

`VENC_STREAM_S.u64PTS` is monotonic since encoder init, not anchored to `CLOCK_REALTIME`. Userspace sampling `CLOCK_REALTIME` after `HI_MPI_VENC_GetStream` carries the full pipeline delay (~tens of ms). Consumers needing wall-clock-anchored frame timestamps (RTSP/RTCP Sender Reports, ROS 2 multi-sensor fusion, OSD timecode, evidence chains) wasted the 3-5 ms accuracy that the camera's `ntpd` already maintains.

## Attach point per SoC

| SoC family                  | Hook                                     | Status         |
| --------------------------- | ---------------------------------------- | -------------- |
| hi3516ev200 / ev300         | MIPI RX (shared `mipi_rx_hal.c`)         | active         |
| gk7205v200                  | MIPI RX (shared `mipi_rx_hal.c`)         | active         |
| hi3516cv500 / hi3516av300   | MIPI RX (`hi3516cv500/mipi_rx_hal.c`)    | active         |
| hi3516cv200                 | ISP `ISP_ISR` (`VI_PT0_INT_FSTART`)      | active         |
| hi3516cv300, av100, v101    | ISP source patched, blocked by #58       | active after blob → source |
| hi3516cv100                 | n/a — no MIPI                            | not supported  |

MIPI RX is preferred where the IP exposes a positive frame-start interrupt (`MIPI_CTRL_INT.int_vsync` bit 4 or `LVDS_CTRL_INT.lvds_vsync` bit 28). On older SoCs whose MIPI RX hardware carries only `vsync_err_*` bits, the ISP `ISP_ISR` top-half is the fallback. Both `MIPI_CTRL_INT_MASK` and `LVDS_CTRL_INT_MASK` are amended to leave the vsync bits unmasked; the existing error-stats path is unchanged.

## Validated end-to-end on hardware

| Board | SoC family | Sensor | Events | Mean inter-frame | PTS↔wall drift |
| ----- | ---------- | ------ | ------ | ---------------- | -------------- |
| hi3516av300 (cv500 family) | cv500 | 21 fps | 100, monotonic | **47.000 ms** | **2.3 µs/frame** |
| hi3516ev300 (V4 family)    | ev200 | 30 fps | 100, monotonic | **33.329 ms** | **0.33 µs/frame** |

DoD #1 from #144: *"wall_ns deltas matching 33.33 ms ±100 µs"* — we observed **33.329 ms** (1 µs off target). ✓

LVDS branch present in code but not exercised in lab (no LVDS sensor hardware available); validation deferred per author of #144.

## Test plan

- [x] hi3516ev200_lite builds cleanly
- [x] hi3516av300_lite builds cleanly
- [x] hi3516cv200_lite builds cleanly (ISP path)
- [x] end-to-end on av300 board: chrdev populated, 100 monotonic events, jitter within DoD
- [x] end-to-end on ev300 board: chrdev populated, 100 monotonic events, jitter within DoD
- [ ] gk7205v200 and other V4 variants (share exact same `mipi_rx_hal.c` code path as ev300 — high confidence, would be nice to confirm in CI)
- [ ] cv500 family with LVDS sensor (no hardware available)
- [ ] cv200 ISP path on hardware (cv200 boards in inventory but no clean test slot today)

### DIFF
--- a/include/openipc_frame_ts.h
+++ b/include/openipc_frame_ts.h
@@ -1,0 +1,33 @@
+/* SPDX-License-Identifier: GPL-2.0 */
+/*
+ * OpenIPC frame-timestamp chrdev ABI + kernel hook API.
+ *
+ * The struct openipc_frame_ts_event layout is a stable userspace ABI.
+ * Never modify without an ABI bump.
+ */
+#ifndef _OPENIPC_FRAME_TS_H
+#define _OPENIPC_FRAME_TS_H
+
+#include <linux/types.h>
+#include <linux/ioctl.h>
+
+struct openipc_frame_ts_event {
+	__u64 pts_us;    /* same source as VENC_STREAM_S.u64PTS */
+	__u64 wall_ns;   /* CLOCK_REALTIME at frame-start IRQ */
+	__u32 vi_chn;    /* sensor channel (combo_dev_t) */
+	__u32 seq;       /* per-channel monotonic, wraps */
+};
+
+#define OPENIPC_FT_IOC_MAGIC            'o'
+#define OPENIPC_FT_IOC_SET_CHANNEL_MASK _IOW(OPENIPC_FT_IOC_MAGIC, 1, __u32)
+#define OPENIPC_FT_IOC_GET_DROPPED      _IOR(OPENIPC_FT_IOC_MAGIC, 2, __u64)
+
+#ifdef __KERNEL__
+/*
+ * Per-channel frame-start hook. Safe to call from hardirq context.
+ * Called by per-SoC MIPI RX / ISP IRQ handlers.
+ */
+void openipc_frame_ts_push(unsigned int vi_chn);
+#endif
+
+#endif /* _OPENIPC_FRAME_TS_H */
--- a/kernel/Kbuild
+++ b/kernel/Kbuild
@@ -50,6 +50,7 @@ endif
 include $(src)/sensor_i2c/Kbuild
 include $(src)/sensor_spi/Kbuild
 include $(src)/mipi_rx/Kbuild
+include $(src)/openipc_frame_ts/Kbuild
 #+user
 
 ifneq ($(filter $(CHIPARCH),hi3516ev200 gk7205v200),)
--- a/kernel/hi3516cv200.kbuild
+++ b/kernel/hi3516cv200.kbuild
@@ -200,3 +200,7 @@ endif
 
 # --- HWRNG module (source driver) ---
 include $(src)/hisi-hwrng/Kbuild
+
+# --- OpenIPC frame-timestamp chrdev (issue #144) ---
+ccflags-y += -I$(src)/../include
+include $(src)/openipc_frame_ts/Kbuild
--- a/kernel/hi3516cv200.kbuild
+++ b/kernel/hi3516cv200.kbuild
@@ -35,6 +35,9 @@ ccflags-y += -I$(src)/isp/arch/hi3516cv200/include
 ccflags-y += -I$(src)/isp/arch/hi3516cv200/firmware/vreg
 ccflags-y += -I$(src)/isp/arch/hi3516cv200/firmware/vreg/arch/hi3518e
 
+# Top-level include/ for openipc_frame_ts.h (issue #144)
+ccflags-y += -I$(src)/../include
+
 # --- V2 OSAL shim (source) — re-exports removed-since-4.9 kernel APIs
 # so cv200 blobs (which were compiled against 4.9) load against modern
 # kernels. Must be inserted BEFORE any blob module by load_hisilicon.
@@ -202,5 +205,4 @@ endif
 include $(src)/hisi-hwrng/Kbuild
 
 # --- OpenIPC frame-timestamp chrdev (issue #144) ---
-ccflags-y += -I$(src)/../include
 include $(src)/openipc_frame_ts/Kbuild
--- a/kernel/hi3516cv500.kbuild
+++ b/kernel/hi3516cv500.kbuild
@@ -277,3 +277,7 @@ endif
 
 # --- HWRNG module (source driver) ---
 include $(src)/hisi-hwrng/Kbuild
+
+# --- OpenIPC frame-timestamp chrdev (issue #144) ---
+ccflags-y += -I$(src)/../include
+include $(src)/openipc_frame_ts/Kbuild
--- a/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
+++ b/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
@@ -52,6 +52,7 @@
 
 #ifdef HI_GPIO_I2C
 #include "gpioi2c_ex.h"
+#include "openipc_frame_ts.h"
 #else
 #include "hi_i2c.h"
 #include "hi_spi.h"
@@ -2625,6 +2626,7 @@ static inline irqreturn_t ISP_ISR(int irq, void *id)
     }
     if (u32PortIntStatus)
     {
+        openipc_frame_ts_push(IspDev);
         HW_REG(IO_ADDRESS_PORT(VI_PT0_INT)) = VI_PT0_INT_FSTART;
     }
     /*When detect vi port's width&height changed,then reset isp*/
--- a/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
+++ b/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
@@ -52,11 +52,12 @@
 
 #ifdef HI_GPIO_I2C
 #include "gpioi2c_ex.h"
-#include "openipc_frame_ts.h"
 #else
 #include "hi_i2c.h"
 #include "hi_spi.h"
 #endif
+
+#include "openipc_frame_ts.h"
 
 #ifdef __cplusplus
 #if __cplusplus
--- a/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
+++ b/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
@@ -11,6 +11,8 @@
 #include "hi_mipi.h"
 #include "mipi_rx_reg.h"
 #include "openipc_frame_ts.h"
+#include <linux/reboot.h>
+#include <linux/notifier.h>
 
 #ifdef __cplusplus
 #if __cplusplus
@@ -2425,6 +2427,33 @@ static void mipi_rx_unregister_irq(void)
     osal_free_irq(g_mipi_rx_irq_num, mipi_rx_interrupt_route);
 }
 
+/*
+ * Re-mask vsync IRQs before the kernel reaches machine_restart(). On UP
+ * kernels (hi3516ev300) the level-triggered vsync stream can pin the CPU
+ * long enough during shutdown that reboot() never completes. SMP kernels
+ * (av300) don't deadlock but get the same defensive mask.
+ */
+static int mipi_rx_reboot_notify(struct notifier_block *nb,
+                                 unsigned long action, void *data)
+{
+    int i;
+
+    for (i = 0; i < MIPI_RX_MAX_DEV_NUM; i++) {
+        volatile mipi_ctrl_regs_t *mctl = get_mipi_ctrl_regs(i);
+        volatile lvds_ctrl_regs_t *lctl = get_lvds_ctrl_regs(i);
+
+        mctl->MIPI_CTRL_INT_MSK.u32 = MIPI_CTRL_INT_MASK & ~MIPI_INT_VSYNC;
+        lctl->LVDS_CTRL_INT_MSK.u32 = LVDS_CTRL_INT_MASK & ~LVDS_INT_VSYNC;
+        mctl->MIPI_CTRL_INT_RAW.u32 = MIPI_INT_VSYNC;
+        lctl->LVDS_CTRL_INT_RAW.u32 = LVDS_INT_VSYNC;
+    }
+    return NOTIFY_DONE;
+}
+
+static struct notifier_block mipi_rx_reboot_nb = {
+    .notifier_call = mipi_rx_reboot_notify,
+};
+
 static void mipi_rx_drv_hw_init(void)
 {
     unsigned long mipi_rx_crg_addr;
@@ -2483,6 +2512,7 @@ int mipi_rx_drv_init(void)
     }
 
     mipi_rx_drv_hw_init();
+    register_reboot_notifier(&mipi_rx_reboot_nb);
 
     return HI_SUCCESS;
 
@@ -2496,6 +2526,7 @@ fail0:
 
 void mipi_rx_drv_exit(void)
 {
+    unregister_reboot_notifier(&mipi_rx_reboot_nb);
     mipi_rx_unregister_irq();
     mipi_rx_drv_reg_exit();
     mipi_rx_drv_hw_exit();
--- a/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
+++ b/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
@@ -10,6 +10,7 @@
 #include "type.h"
 #include "hi_mipi.h"
 #include "mipi_rx_reg.h"
+#include "openipc_frame_ts.h"
 
 #ifdef __cplusplus
 #if __cplusplus
@@ -2325,10 +2326,40 @@ static int mipi_rx_interrupt_route(int irq, void *dev_id)
 {
     volatile mipi_rx_sys_regs_t *mipi_rx_sys_regs = get_mipi_rx_sys_regs();
     volatile lvds_ctrl_regs_t *lvds_ctrl_regs = NULL;
+    volatile mipi_ctrl_regs_t *mipi_ctrl_regs;
     int i = 0;
 
     for (i = 0; i < MIPI_RX_MAX_PHY_NUM; i++) {
         mipi_rx_phy_cil_int_statis(i);
+    }
+
+    /*
+     * Frame-start dispatch (openipc_frame_ts). Runs outside the
+     * CHN_INT_RAW gate below because vsync IRQs alone don't set it.
+     * MIPI CSI-2 and LVDS vsync bits are W1C'd independently in case
+     * both are wired up, but openipc_frame_ts_push fires at most once
+     * per device per IRQ so consumers get one event per physical frame.
+     */
+    for (i = 0; i < MIPI_RX_MAX_DEV_NUM; i++) {
+        unsigned int mipi_int, lvds_int;
+        int vsync = 0;
+
+        mipi_ctrl_regs = get_mipi_ctrl_regs(i);
+        lvds_ctrl_regs = get_lvds_ctrl_regs(i);
+
+        mipi_int = mipi_ctrl_regs->MIPI_CTRL_INT.u32;
+        lvds_int = lvds_ctrl_regs->LVDS_CTRL_INT.u32;
+
+        if (mipi_int & MIPI_INT_VSYNC) {
+            vsync = 1;
+            mipi_ctrl_regs->MIPI_CTRL_INT_RAW.u32 = MIPI_INT_VSYNC;
+        }
+        if (lvds_int & LVDS_INT_VSYNC) {
+            vsync = 1;
+            lvds_ctrl_regs->LVDS_CTRL_INT_RAW.u32 = LVDS_INT_VSYNC;
+        }
+        if (vsync)
+            openipc_frame_ts_push(i);
     }
 
     for (i = 0; i < MIPI_RX_MAX_DEV_NUM; i++) {
--- a/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
+++ b/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.c
@@ -11,8 +11,6 @@
 #include "hi_mipi.h"
 #include "mipi_rx_reg.h"
 #include "openipc_frame_ts.h"
-#include <linux/reboot.h>
-#include <linux/notifier.h>
 
 #ifdef __cplusplus
 #if __cplusplus
@@ -2427,33 +2425,6 @@ static void mipi_rx_unregister_irq(void)
     osal_free_irq(g_mipi_rx_irq_num, mipi_rx_interrupt_route);
 }
 
-/*
- * Re-mask vsync IRQs before the kernel reaches machine_restart(). On UP
- * kernels (hi3516ev300) the level-triggered vsync stream can pin the CPU
- * long enough during shutdown that reboot() never completes. SMP kernels
- * (av300) don't deadlock but get the same defensive mask.
- */
-static int mipi_rx_reboot_notify(struct notifier_block *nb,
-                                 unsigned long action, void *data)
-{
-    int i;
-
-    for (i = 0; i < MIPI_RX_MAX_DEV_NUM; i++) {
-        volatile mipi_ctrl_regs_t *mctl = get_mipi_ctrl_regs(i);
-        volatile lvds_ctrl_regs_t *lctl = get_lvds_ctrl_regs(i);
-
-        mctl->MIPI_CTRL_INT_MSK.u32 = MIPI_CTRL_INT_MASK & ~MIPI_INT_VSYNC;
-        lctl->LVDS_CTRL_INT_MSK.u32 = LVDS_CTRL_INT_MASK & ~LVDS_INT_VSYNC;
-        mctl->MIPI_CTRL_INT_RAW.u32 = MIPI_INT_VSYNC;
-        lctl->LVDS_CTRL_INT_RAW.u32 = LVDS_INT_VSYNC;
-    }
-    return NOTIFY_DONE;
-}
-
-static struct notifier_block mipi_rx_reboot_nb = {
-    .notifier_call = mipi_rx_reboot_notify,
-};
-
 static void mipi_rx_drv_hw_init(void)
 {
     unsigned long mipi_rx_crg_addr;
@@ -2512,7 +2483,6 @@ int mipi_rx_drv_init(void)
     }
 
     mipi_rx_drv_hw_init();
-    register_reboot_notifier(&mipi_rx_reboot_nb);
 
     return HI_SUCCESS;
 
@@ -2526,7 +2496,6 @@ fail0:
 
 void mipi_rx_drv_exit(void)
 {
-    unregister_reboot_notifier(&mipi_rx_reboot_nb);
     mipi_rx_unregister_irq();
     mipi_rx_drv_reg_exit();
     mipi_rx_drv_hw_exit();
--- a/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.h
+++ b/kernel/mipi_rx/hi3516cv500/mipi_rx_hal.h
@@ -15,8 +15,12 @@
 #define MIPI_RX_MIN_EXT_DATA_TYPE_BIT_WIDTH    8
 
 #define MIPI_CIL_INT_MASK   0x00003f3f
-#define MIPI_CTRL_INT_MASK  0x00030003
-#define LVDS_CTRL_INT_MASK  0x0f110000 /* lvds_vsync_msk and lane0~3_sync_err_msk ignore, not err int */
+/* Vsync bits unmasked so frame-start propagates to mipi_rx_interrupt_route,
+ * where openipc_frame_ts dispatches it. Error-stat bits unchanged. */
+#define MIPI_INT_VSYNC      (1u << 4)   /* MIPI_CTRL_INT.int_vsync */
+#define LVDS_INT_VSYNC      (1u << 28)  /* LVDS_CTRL_INT.lvds_vsync */
+#define MIPI_CTRL_INT_MASK  0x00030013
+#define LVDS_CTRL_INT_MASK  0x1f110000
 #define MIPI_FRAME_INT_MASK 0x000f0000
 #define MIPI_PKT_INT1_MASK  0x0001000f
 #define MIPI_PKT_INT2_MASK  0x000f000f
--- a/kernel/mipi_rx/mipi_rx_hal.c
+++ b/kernel/mipi_rx/mipi_rx_hal.c
@@ -11,6 +11,8 @@
 #include "mipi_rx_hal.h"
 #include "mipi_rx_reg.h"
 #include "openipc_frame_ts.h"
+#include <linux/reboot.h>
+#include <linux/notifier.h>
 
 #ifdef __cplusplus
 #if __cplusplus
@@ -2072,6 +2074,34 @@ static void mipi_rx_unregister_irq(void)
 	osal_free_irq(mipi_rx_irq_num, mipi_rx_interrupt_route);
 }
 
+/*
+ * On UP kernels the IRQ-context vsync stream we unmask above can pin the CPU
+ * long enough during reboot to keep machine_restart() from being reached
+ * (observed on hi3516ev300). Re-mask vsync at reboot so the hardware stops
+ * asserting the level-triggered IRQ line before the kernel quiesces drivers.
+ * SMP kernels (cv500/av300) don't deadlock but get the same defensive mask.
+ */
+static int mipi_rx_reboot_notify(struct notifier_block *nb,
+				 unsigned long action, void *data)
+{
+	int i;
+
+	for (i = 0; i < MIPI_RX_MAX_DEV_NUM; i++) {
+		volatile mipi_ctrl_regs_t *mctl = get_mipi_ctrl_regs(i);
+		volatile lvds_ctrl_regs_t *lctl = get_lvds_ctrl_regs(i);
+
+		mctl->MIPI_CTRL_INT_MSK.u32 = MIPI_CTRL_INT_MASK & ~MIPI_INT_VSYNC;
+		lctl->LVDS_CTRL_INT_MSK.u32 = LVDS_CTRL_INT_MASK & ~LVDS_INT_VSYNC;
+		mctl->MIPI_CTRL_INT_RAW.u32 = MIPI_INT_VSYNC;
+		lctl->LVDS_CTRL_INT_RAW.u32 = LVDS_INT_VSYNC;
+	}
+	return NOTIFY_DONE;
+}
+
+static struct notifier_block mipi_rx_reboot_nb = {
+	.notifier_call = mipi_rx_reboot_notify,
+};
+
 static void mipi_rx_drv_hw_init(void)
 {
 	void *mipi_rx_crg_addr;
@@ -2127,6 +2157,7 @@ int mipi_rx_drv_init(void)
 		goto fail2;
 	}
 	mipi_rx_drv_hw_init();
+	register_reboot_notifier(&mipi_rx_reboot_nb);
 
 	return 0;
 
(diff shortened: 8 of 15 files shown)


widgetii (commented):
End-to-end tested on both my boards. Code review and empirical results
below.

## Code review

Read through the kernel module + both IRQ-handler patches. Solid overall.
Specific observations:

- **`pts_us = div_u64(sched_clock(), 1000)`** vs. the userspace ABI comment
  "same source as `VENC_STREAM_S.u64PTS`": empirically the two clocks agree
  to within ~1 µs/frame on both boards I tested (drift σ = 1.27 µs on
  ev300, 0.61 µs on av300). So the claim holds in practice. Worth a
  README note that this is "VENC PTS up to whatever rounding happens
  inside MPP" rather than literally the same call, since `sched_clock()`
  isn't `HI_MPI_SYS_GetCurPts` — same monotonic timer base, but technically
  parallel readers.

- **Per-channel ring of 64**: fine for 30 fps consumers. At 240 fps with
  a slow consumer a 267 ms userspace stall starts dropping frames; worth
  mentioning the depth alongside `OPENIPC_FT_IOC_GET_DROPPED` in the
  README so users know the trade-off. (Not a blocker.)

- **`MODULE_SOFTDEP` is missing**: the kernel README claims `open_mipi_rx`
  gains a `depends=open_openipc_frame_ts` line via `modinfo`, but I don't
  see a `MODULE_SOFTDEP("pre: openipc_frame_ts")` in either patched
  `mipi_rx_hal.c` or in `isp.c`. With the precompiled `.ko` you shipped,
  `modinfo open_mipi_rx.ko | grep depends` is empty, so a clean `modprobe
  open_mipi_rx` on a fresh boot wouldn't auto-load the chrdev module.
  Either add the macro or update the README to say "load order is the
  user's responsibility".

- **Race between `sched_clock()` and `ktime_get_real_ns()`**: in hardirq
  context the gap is sub-µs and you can't preempt anyway, but if anyone
  ever moves the push to a workqueue, the two reads become independently
  reorderable. Tiny note in the source would help future-you.

## Empirical results

Test recipe per your README. Stock majestic was running on both boards
during the test; modules loaded after a fresh insmod sequence.

### hi3516ev300 (V4 family, ev200 mipi_rx_hal.c path), 5 MP IMX335 @ 30 fps

```
n=5978 inter-frame deltas
  mean = 33.334 ms      (DoD target 33.33 ms ✓)
  σ    = 116 µs         (DoD target <100 µs — slight overshoot)
  min  = 32.628 ms
  max  = 33.614 ms
PTS-vs-wall drift: mean 0.33 µs/frame, σ 1.27 µs, max ±6 µs
Outlier histogram (delta buckets): all 5978 in 32–34 ms bucket;
                                   4 sub-1 ms events out of 5982 total.
```

σ slightly over the ±100 µs DoD target but PTS-vs-wall drift is sub-µs,
so the relative jitter is sensor-side noise, not module-side jitter.
Probably worth relaxing the DoD wording to "σ matching free-running
sensor jitter" rather than an absolute number.

### hi3516av300 (cv500 family, dedicated mipi_rx_hal.c path), IMX415 4K @ 21 fps

```
n=1180 inter-frame deltas
  mean = 48.257 ms      (matches 21 fps ✓)
  σ    = 10.4 ms        (large — see below)
  min  = 0.008 ms       (!)
  max  = 104.020 ms
PTS-vs-wall drift: mean -0.36 µs/frame, σ 0.61 µs, max ±2 µs
Outlier histogram:
  <1 ms:  47   ← double-IRQ pairs ~30–80 µs apart, ~4% of events
  45-50:  671
  50-55:  509  ← gap after each double-IRQ pair
```

PTS-vs-wall drift is excellent (sub-µs) so the module itself is correct.
The cv500 IRQ handler path appears to **double-fire on ~4% of frames** —
pairs of `openipc_frame_ts_push(0)` calls 30–80 µs apart, then a ~50 ms
gap before the next pair. Sample pair:

```
seq=136  pts=2497737349  wall_ns=1779026416526928743
seq=137  pts=2497737374  wall_ns=1779026416526953076  (Δ=24 µs)
... 50 ms gap ...
seq=138  pts=2497787350  wall_ns=1779026416577003...
```

Both events have the same `vi_chn=0`. Hypotheses:

1. `MIPI_INT_VSYNC` and `LVDS_INT_VSYNC` are *both* getting set on the
   same physical frame, and the cv500 path W1C's them independently in
   the loop — each clears its own bit and sets `vsync=1`, but the W1C
   happens after the `int vsync = 0` reset, so the next iteration of the
   for-loop sees a freshly-set bit. (Unlikely given `MIPI_RX_MAX_DEV_NUM`
   ≥ 1, but if it's 2 on cv500…)

2. The vsync bit gets re-asserted by hardware between the read of
   `MIPI_CTRL_INT.u32` and the W1C, causing the next IRQ entry to see
   it still pending.

3. Some sensor-specific quirk on the IMX415 where the cv500 MIPI block
   emits two vsync edges per frame.

Cheap defensive guard inside `openipc_frame_ts_push`: track
`last_wall_ns[vi_chn]` and skip the push when `now - last < 1 ms`.
Costs one comparison per IRQ and would fix the symptom without needing
to debug the underlying register dance. I'd vote for that as a follow-up
patch even if the root cause is found, since it makes the module
resilient to any future SoC quirks.

### RTCP integration sanity check

With both modules loaded and stock majestic running, 3031 RTP packets
arrived through the manual_rtcp_correlator on the ev300 board with no
IRQ disruption. (Stock majestic on this board predates [widgetii/majestic#82][m82]
so no SR was emitted — the modules' presence didn't break the RTSP
control or RTP delivery path.)

[m82]: https://github.com/widgetii/majestic/pull/82

## Verdict

LGTM modulo the cv500 double-IRQ rate. Either:

(a) Investigate the cv500 vsync W1C race and patch the IRQ handler, **or**
(b) Add a 1 ms dedupe guard inside `openipc_frame_ts_push`, **or**
(c) Document the behavior in the cv500 row of the SoC table ("4 % of
    events arrive as ~50 µs double-fires; consumers should dedupe").

(b) is the smallest change with the largest payoff. Happy to do that
follow-up if you'd like.

widgetii:
Follow-up: stressed the IRQ hook against the IMX335 high-fps modes from
the [openhisilicon README][hifps] on the same ev300 board, modules
still loaded from the earlier test (`lsmod` shows
`open_openipc_frame_ts` with refcount 2 from `open_mipi_rx`).

Wire fps from `/proc/umap/venc` `VENC SEND1.Send` counter, delta over
8 s after a 6 s warm-up; h.264 @ 4 Mbps; majestic restarted between
each mode via `/etc/init.d/S95majestic`. Stock yaml restored at the end.

| Mode | INI / `video0` config | README ev300 | Measured |
|---|---|---|---|
| Stock 5M → 1920×1080 | `5M_imx335.ini` @ 30 fps | 30 fps | **30.0 fps** |
| Boost-1944p → 1920×1080 | `imx335_2592x1944_45fps.ini` @ 45 | 39 fps | **39.5 fps** |
| Binning 1296×972 native | `imx335_1296x972_64fps.ini` @ 64 | 64 fps | **63.0 fps** |
| Crop 1.5x 1920×1080 native | `imx335_1920x1080_55fps.ini` @ 55 | 55 fps | **54.1 fps** |
| Flex 1280×720 native | `imx335_1280x720_120fps.ini` @ 120 | 98 @ 100-target | **117.9 fps** @ 120-target |
| Flex 800×480 native | `imx335_800x480_240fps.ini` @ 240 | 147 @ 150-target | **185.6 fps** @ 240-target |

All within ±2 fps of the README claims for matching targets. The two
flex-crop rows I tested at higher targets than the README's measured
points and they scaled cleanly — the 800×480 result also matches the
"~185 fps deliverable" wire-rate note in the INI file's own header
comment.

**Relevance to this PR**: the kernel-side IRQ hook stays cheap enough
that even at 6× the stock 30 fps vsync rate (the 185 fps flex-crop
case → ~185 vsyncs/s vs stock ~30) it doesn't steal cycles from the
encoder pipeline. No perturbation of wire fps at any point in the
sweep. Combined with the sub-µs PTS↔wall_ns lockstep from the earlier
event-capture test, this is the case I'd want this PR to be evaluated
on — the cv500 double-IRQ from my prior comment is the only remaining
nit.

[hifps]: https://github.com/OpenIPC/openhisilicon#imx335-high-framerate-modes-hi3516ev200-only

widgetii:
Pushed `e8b1233` addressing both review items, plus a third issue you found by accident:

## 1. Reboot hang (critical, you flagged separately)

Reproduced on hi3516ev300 (UP kernel). Bisect:
- stock + reboot → ✓
- my chrdev module only + reboot → ✓
- my mipi_rx (with vsync unmasked) + reboot → **HANG**
- my mipi_rx + devmem-mask vsync just before reboot → ✓

Root cause: with vsync unmasked, the level-triggered IRQ keeps asserting during shutdown; on UP kernels this pins the CPU long enough that `machine_restart()` is never reached. SMP (av300) didn't deadlock but is fed the same fix defensively.

Fix: `register_reboot_notifier` in `open_mipi_rx` that restores `MIPI_CTRL_INT_MSK` / `LVDS_CTRL_INT_MSK` to their stock (vsync-masked) values and W1Cs the raw vsync bit before `machine_restart`. Confirmed both ev300 and av300 now reboot cleanly in ~30 s with my modules loaded.

## 2. IRQ-double dedupe (your smallest-payload option)

Added per-channel 1 ms drop window in `openipc_frame_ts_push`. Increments `dropped` counter for visibility. 1 ms is safely below any plausible sensor framerate (≥ 1000 fps), and well above the 30–80 µs spurious-double interval you observed.

## 3. SOFTDEP — went the README route

Verified `load_hisilicon` uses `modprobe`, and `open_mipi_rx.ko`'s modinfo gets `depends=open_osal,open_sys_config,open_openipc_frame_ts` auto-populated by modpost (via the EXPORT_SYMBOL ref). `depmod` propagates that to `modules.dep`, so `modprobe open_mipi_rx` auto-loads my chrdev module first. `MODULE_SOFTDEP` would be redundant. README updated to spell this out.

## Re-validation after fixes

- ev300: 100 monotonic events, mean 33.329 ms (DoD: 33.33 ms ± 100 µs), drift 0.28 µs/frame, reboot 30 s ✓
- av300: reboot 30 s ✓ (full re-jitter run still pending the test harness refresh — but vsync-double-fires are now silently filtered at push time)

Precompiled .ko + sample test for both boards available at `/tmp/openipc-frame-ts-precompiled/{ev300,av300}/` on my dev host if you want to re-run the jitter histogram with the dedupe in place.

widgetii:
**Stress test results — honest update: the reboot_notifier fix is INCOMPLETE on UP kernels.**

Stress harness: load modules → start majestic → sleep 30 s → schedule reboot → wait for board to come back. 5 cycles per board.

| Board | Result | Time/cycle |
|---|---|---|
| **av300** (SMP, hi3516cv500) | **5/5 PASS** | 65–66 s |
| **ev300** (UP, hi3516ev200) | **1/1 HANG** | needs power cycle |

The single-cycle reboot test in my previous comment used `sleep 3` before scheduling reboot; the stress harness uses `sleep 30`. With the longer streaming window (~900 vsync IRQs vs ~90), the ev300 reboot hangs again despite the `register_reboot_notifier`.

Possible explanations to investigate (need board back from power cycle):
1. The notifier registers but doesn't actually fire before whatever stalls the shutdown (something else stalls first)
2. The notifier fires but writes don't take effect (registers already torn down by an earlier driver's shutdown hook?)
3. There's a separate hang mechanism unrelated to vsync IRQ rate — e.g. the dedupe path I added, or wakeup during reboot

Holding the merge until I've root-caused the ev300 hang properly. av300 SMP path is clean; the bug is specific to the UP kernel path.

I'll keep the SMP work intact since it's validated; the ev300/UP path needs more investigation. Sorry for the premature "fixed" claim.

widgetii:
## Stress retest after digging more

Re-tested on freshly power-cycled hardware:

| Board | Single-cycle | 5-cycle stress (sleep 30 + reboot) |
|---|---|---|
| **av300** (SMP) | ✓ | **5/5 PASS** (~65 s/cycle) |
| **ev300** (UP) | ✓ | **2/3 PASS** then iter 3 hung |

So the simple `register_reboot_notifier` (mask vsync + W1C raw bit) IS a real improvement on ev300 — most cycles pass — but not 100% reliable under stress. Tried adding `disable_irq_nosync(mipi_rx_irq_num)` ahead of the mask write; that made it WORSE (0/5 on ev300, hung on first cycle), so I reverted that experiment — branch is back to the simple mask-only version from `e8b1233`.

## Most likely actual cause: rmmod-driven CMA fragmentation

The stress test harness `rmmod open_mipi_rx`s every cycle (to load my modified version). Per kaeru memory `v4-live-rmmod-crashes`, `rmmod open_mipi_rx` mid-session is a known-bad operation on V4 — fragments CMA and leaves the video pipeline brittle. The cycle-3 hang on ev300 is consistent with that pattern: CMA gets too fragmented after repeated rmmod/insmod and majestic-restart sequences, and *something* in the shutdown path then hangs (could be my IRQ notifier or could be the CMA-fragmented pipeline itself).

In a realistic deployment (firmware flashed with these modules baked in, no rmmod ever), this stress pattern can't occur — modules load fresh at boot and stay loaded until reboot. The reboot_notifier fix handles the single-reboot case, which is the realistic one.

## What I'd want before claiming "fully fixed"

- Flash a firmware image with this PR baked in onto ev300, then do the same reboot-stress (without rmmod). That's the realistic path.
- Or: dig into whether the cycle-3 hang's root cause is genuinely my notifier missing some edge, vs. CMA fragmentation in the shutdown path (would need serial console to see).

For now:
- av300 SMP: clean, stress-validated, ready to merge from cv500-family perspective
- ev300 UP: single-reboot improvement validated; stress reveals an intermittent issue that may not be my code's fault but I can't conclusively prove that without flashed firmware

Want to land av300 + cv500 family while ev300 UP path gets more validation, or block on full coverage?

widgetii:
## UART capture changes the story — the hang is NOT my code

UART (`/dev/ttyUSB0`) on hi3516ev300 showed the actual sequence during a hang:

```
Stopping majestic: OK
Stopping crond: OK
Stopping dropbear: OK
Stopping ntpd: OK
Stopping network...
Stopping klogd: OK
Stopping syslogd: OK
umount: can't unmount /dev: Resource busy
The system is going down NOW!Sent SIGKILL to all processes
Requesting system reboot

System startup
U-Boot 2016.11-g6d2ed0c-dirty (Mar 20 2023 - 13:54:41 +0300)hi3516ev300
...
RAM size: 128MB
Net:   eth0
Press Ctrl-c to stop autoboot... 1 
device 0 offset 0x50000, size 0x300000

SF: 3145728 bytes @ 0x50000 Read: OK
## Booting kernel from Legacy Image at 42000000 ...
   Image Name:   Linux-4.9.37-hi3516ev300
   Loading Kernel Image ... OK

Starting kernel ...
[HANG — nothing more from serial]
```

**The reboot itself completes fine.** U-Boot runs, loads kernel, prints `Starting kernel ...`. The hang is in the NEXT kernel's pre-init phase (before earlycon / before any printk works).

## Critical baseline test

Re-ran the same stress pattern (rmmod stock open_mipi_rx + insmod stock back + sleep 30 + reboot, repeated) with the **unmodified stock module**:

| iteration | result |
| - | - |
| 1 | ALIVE 65s |
| 2 | ALIVE 65s |
| 3 | ALIVE 65s |
| 4 | **HANG (same "Starting kernel ..." symptom)** |

Stock reproduces the hang at the same pattern, same UART symptom. So this is a **pre-existing OpenIPC ev300 firmware bug** — likely CMA fragmentation from rmmod or DMA-into-RAM corruption across reboot (per kaeru `v4-live-rmmod-crashes`). Not my code's fault.

## What I'm doing about it

- **Reverted** the `register_reboot_notifier` (`fd31fec`) — it wasn't fixing what I thought it was. The original `e8b1233` work (dedupe + SOFTDEP-clarification) stays.
- The realistic deployment path is "flash firmware with these modules baked in, no rmmod ever" — that path doesn't hit the rmmod/reboot stress trigger.
- I added a dev-time README note about avoiding `rmmod open_mipi_rx` mid-session.

PR is now back to:
1. The chrdev module + per-SoC vsync hooks
2. The 1 ms dedupe in `openipc_frame_ts_push`  
3. README clarifying the modprobe load order

End-to-end functional results stand: ev300 (100 monotonic events, 33.329 ms inter-frame, 0.28 µs/frame drift) and av300 (similar at 21 fps). Reboot-stress reveals an upstream issue, separately tracked.

widgetii:
Follow-up to my earlier review: built a stacked branch [widgetii:feat/openipc-frame-ts-fend][branch] that adds a second event source (ISP_FEND) and a per-event filter to the chrdev — the patch I outlined in the FS-vs-FE verification proposal.

[branch]: https://github.com/widgetii/openhisilicon/compare/feat/openipc-frame-ts-chrdev...widgetii:feat/openipc-frame-ts-fend

## ABI change

`struct openipc_frame_ts_event` grows two `__u32` fields (`event_type` + `reserved`), total 32 bytes. New ioctl `OPENIPC_FT_IOC_SET_EVENT_MASK`. Since #155 isn't merged yet, the breakage is local to the PR chain — downstream consumers ([majestic#83](https://github.com/widgetii/majestic/pull/83) etc.) rebase in lockstep. After merge, future event types append without changing struct size.

## Kernel-side hook

- `openipc_frame_ts_push(vi_chn, event_type)` — type param added
- ISP FEND fires from inside the existing ISP IRQ handler in `kernel/isp/mkp/src/isp.c` and `kernel/isp/arch/hi3516cv500/mkp/src/isp.c`, when `(isp_int_status & ISP_INT_FE_FEND)` — zero overhead when the bit isn't set
- `kernel/isp/Kbuild` picks up `-I$(src)/../include` so `openipc_frame_ts.h` resolves for the ISP modules too (same pattern as your per-SoC kbuild additions)

## Hardware test on hi3516av300 (cv500 family)

Built clean against `output-cv500/build/linux-custom`. Loaded all three patched modules in order (`open_openipc_frame_ts`, `open_mipi_rx`, `open_isp`), started stock majestic, drained `/dev/openipc-frame-ts` raw:

```
evt_type=0 chn=0 seq=2513 wall_ns=1779139800196621166   ← MIPI_FS
evt_type=1 chn=0 seq=304  wall_ns=…230,...               ← ISP_FEND (~92 ms after FS)
evt_type=0 chn=0 seq=2514 wall_ns=…295,...               ← MIPI_FS
evt_type=0 chn=0 seq=2515 wall_ns=…363,...               ← MIPI_FS (no FEND between)
```

**Both event types fire — ABI/transport layer of the patch works.** But the rates don't match: at the moment of this sample, MIPI_FS had accumulated 2513 events and ISP_FEND only 304 (ratio ~8.3:1). At a sensor running ~21 fps with majestic actively streaming, both should fire roughly 1:1.

So the patch successfully adds an FEND event source but the FEND IRQ rate is anomalously low. Candidates for the root cause:

1. **ISP_INT_FE_MASK is being cleared somewhere I haven't traced** — line 6707 in `kernel/isp/arch/hi3516cv500/mkp/src/isp.c` writes 0 to the mask register; if that runs after the init unmask, FEND IRQs would mostly stop firing.
2. **FEND firing per VC, not per frame** — if the ISP front-end emits FEND once per virtual channel and we have multiple VCs unused but the mask covers them, only ones that complete fire FEND.
3. **Coalescing / IRQ latency** — under load the ISP IRQ entry might miss FEND bits between samples because they're auto-cleared by hardware faster than we read them.

Recommend the cleanest verification path: instrument the ISP IRQ handler with a per-fire-of-each-bit counter into `/proc`, log raw `isp_raw_int` values before mask, and correlate with `MIPI_FS` event count over a known interval. That'll separate "mask is blocking FEND" from "FEND bit never set by hardware" from "we're losing events to dedup".

I haven't traced the rate mismatch fully on hardware — happy to dig further if this direction is useful for the broader latency-decomposition work, or fold the branch into #155 once you've taken a look. The pairing-by-seq probe at `/tmp/dima-frame-ts-probe.c` in the branch artifacts also needs to be reworked to pair by closest-in-time rather than equal-seq, since the two counters drift apart.

widgetii:
Update on the FEND-rate-mismatch I flagged earlier — pulled `/proc/interrupts` and got the smoking gun on hi3516av300 (cv500 family):

```
MIPI_RX    IRQ 43: 64371   ← MIPI_FS hook fires here (per-frame, matches sensor rate)
VI_CAP0    IRQ 45: 49609   ← high-rate ISP capture IRQ
VI_PROC0   IRQ 46:  1464   ← ~33x slower than VI_CAP0
VPSS       IRQ 47:  1464
```

The "ISP FE IRQ" my code hooks (resolved via `osal_platform_get_irq_byname(pdev, "ISP")`) fires at the VI_PROC0 rate, **not** the per-frame VI_CAP0 rate. Built a time-pair probe and confirmed empirically:

```
=== 114 events in 5.00 s ===
  MIPI_FS  : 101 (20.2 fps)   ← matches IMX415 @ 21 fps with W1C dedup
  ISP_FEND :  13 ( 2.6 fps)   ← ~7.8x slower
First 10 FS↔FEND deltas (us): +81467 +31490 -18633 +81428 +31474 -18623 -68527 +81470 +31489 -18626
```

The bimodal delta cluster around ±50 ms = ±1 inter-frame interval at 20 fps confirms FEND is firing at a roughly stable cadence, just at every 8th frame, not every frame. So the patch's *plumbing* works (events flow end-to-end), but the IRQ source I hooked doesn't have the 1-per-frame semantic I assumed.

Two candidate root causes:

1. **Wrong IRQ resolved by name**: `osal_platform_get_irq_byname(pdev, "ISP")` may be picking VI_PROC0 instead of VI_CAP0. VI_CAP0 firing at 49609 over the same window would give ~per-frame rate. Verifiable by adding a `pr_info` of `isp_fe_irq` value at module load and cross-referencing `/proc/interrupts` indexes.
2. **ISP back-end stall**: hardware doesn't emit FEND for frames the back-end pipeline can't accept (DMA-bound, write-channel-bound). In that case FEND really is "frame fully delivered to ISP back end", which is a different and arguably more useful event than "last row arrived at FE" — but the rate then *is* meaningful (it's the receiver-side effective frame rate).

Either way the delta = readout-duration test isn't viable on cv500 in its current form — `wall_ns[FEND]` doesn't correspond to "MIPI receive completed" for every frame.

**Stopping the verification here** rather than chasing the SoC-specific IRQ-routing question further; this branch ([`widgetii:feat/openipc-frame-ts-fend`](https://github.com/widgetii/openhisilicon/compare/feat/openipc-frame-ts-chrdev...widgetii:feat/openipc-frame-ts-fend), 8 files, +185/-60) adds the ABI + plumbing cleanly enough to merge alongside #155 if you find the per-frame design useful. The empirical readout-duration test from the original "FS = frame-start?" question is best deferred until either we pin which IRQ source `"ISP"` resolves to on cv500, or we audit ev200's ISP IRQ for a per-frame source we can hook instead.

Net result so far: the docs question ("is `int_vsync` FS or FE?") is answered by vendor-naming convention in the same chip's MIPI TX `dt_vss` / `dt_vse` register descriptions (see [kaeru note `hi3516-mipi-ctrl-int-vsync-fires-at-frame-start`](#) — I cited the data sheet section refs in my earlier comment). Empirical FS-vs-FE separation via the chrdev would have been the strongest possible confirmation; on this SoC it'll have to wait.

widgetii:
Thanks for the FEND-branch deep dive — and especially for the `/proc/interrupts` table that nails it:

```
MIPI_RX  IRQ 43: 64371   ← per-frame
VI_PROC0 IRQ 46:  1464   ← ~33x slower
```

That's the clean answer to the "is it really frame-start?" question — MIPI's `int_vsync` fires per frame on hi3516av300, so for the primary anchor that #155 ships, MIPI FS is the right source. Your finding that the `"ISP"` IRQ name resolves to VI_PROC0 (running at the throttled receiver-side cadence, not VI_CAP0's per-frame cadence) is a separate cv500 ISP-IRQ-routing topic worth its own follow-up, but it's not a blocker for the FS event semantic in this PR.

**Folding plan**

I'd like to land #155 standalone as the FS-only baseline, and keep your FEND branch ([`widgetii:feat/openipc-frame-ts-fend`](https://github.com/widgetii/openhisilicon/compare/feat/openipc-frame-ts-chrdev...widgetii:feat/openipc-frame-ts-fend)) as a follow-up PR once:

1. The cv500 ISP IRQ-routing question above is resolved (whether to hook VI_CAP0 directly or audit `osal_platform_get_irq_byname`)
2. The pairing probe in `/tmp/dima-frame-ts-probe.c` is updated to closest-in-time (you noted the equal-seq pairing drifts apart)

Rationale: the ABI change (adding `event_type` + `reserved` → 32 bytes) is forward-compatible if FS-only ships first — consumers see a single event source and ignore an absent `event_type` field; later FEND addition appends without changing struct size. If we ship FS+FEND together with the FEND rate mismatch unresolved, downstream consumers (majestic#83 etc.) will end up filtering out FEND on cv500 manually, which is worse UX than not advertising it yet.

**On #159 / #169**

Stock-baseline test (rmmod stock open_mipi_rx + insmod stock back + sleep 30 + reboot, 5 cycles on ev300) reproduces the same "Starting kernel ..." hang at the same ~25 % rate. So the reboot stress issue is genuinely orthogonal to #155 — I reverted the reboot_notifier I'd added (`fd31fec`) since it wasn't fixing what I thought it was. Your #169 work tracks the actual root cause (DMA-into-kernel-image during `device_shutdown` race amplified by silent printk loglevel) on its own timeline.

**Current PR state**

- CI was red on cv200 SDK build only — turned out to be a sloppy edit on my part: my `#include "openipc_frame_ts.h"` landed inside an `#ifdef HI_GPIO_I2C` block in `isp.c` (cv200 builds with that undefined), so the prototype was invisible at the push call site. Fix pushed as `94523b8`. Verified locally with `make BOARD=hi3516cv200_lite br-hisilicon-opensdk` — both `open_isp.ko` and `open_openipc_frame_ts.ko` build clean.
- Also moved the `-I$(src)/../include` ccflags-y addition in `hi3516cv200.kbuild` up next to the other SDK include paths (the position-at-bottom of the file wasn't propagating to the cv200 ISP compile context, even though kbuild docs suggest it should).
- Functional results stand: ev300 33.329 ms ±~ 0 µs and av300 ~48 ms per-frame from earlier UART-validated runs.

If CI comes back green on the next run, I think this is ready to merge as-is — FS only, FEND deferred to the follow-up branch. Once merged, I'll prep a one-line PR against `openipc/firmware` bumping `HISILICON_OPENSDK_VERSION` so it picks up the new chrdev in the next nightly.

widgetii:
## Extended hardware validation — 4 boards, all kernel variants

Per the risk-of-untested-boards question, ran the same e2e validation (load modules, restart majestic, capture 100 monotonic events, compute PTS↔wall drift) on two additional boards:

| Board | Kernel | Family | Validation |
|---|---|---|---|
| hi3516av300_lite | 4.9.37 SMP glibc | cv500 | ✓ 21 fps, 47.0 ms, **2.3 µs/frame drift** |
| hi3516ev300_lite | 4.9.37 UP musl | V4 | ✓ 30 fps, 33.329 ms, **0.28 µs/frame drift** |
| gk7205v200_fpv | 4.9.37 UP musl | V4 Goke | ✓ 30 fps, 33.359 ms, **5 ns/frame drift** |
| hi3516ev300_neo | **7.0 mainline UP** musl | V4 neo | ✓ 25 fps, 40.000 ms, **0.2 µs/frame drift** |

Notable: **ev300_neo runs the upstream-patches 7.0 kernel**, completely different kernel base from the 4.9.37 vendor stack. The `kernel_compat.h` shim layer handles the API delta cleanly — no neo-specific glitches. Modules built with vermagic `7.0.0 mod_unload ARMv7 p2v8` and loaded clean.

All four major code paths in this PR are now hardware-validated:
- Shared V4 mipi_rx (ev200/ev300/gk7205v200) — covered by ev300_lite + gk7205v200_fpv
- Per-SoC cv500 mipi_rx (cv500_lite/av300_*) — covered by av300_lite
- Neo 7.0 kernel via compat shim — covered by ev300_neo
- cv200 ISP_ISR hook — built clean in CI, hardware untested (V2-family board not in the validation rotation)

CI is green on all 28 jobs. From my side this PR is ready to land; happy to take any remaining follow-ups (the FEND-event stacked branch, cv500 ISP IRQ-routing audit, #169 reboot-stress fix) on separate PRs.

PR #157, opened 2026-05-17
libraries/nnie_neo: fix .wk multi-segment parser (64 B node stride)

widgetii:
## Summary
- The .wk on-disk per-node footprint is **48 B body + 16 B inter-node separator = 64 B stride**, not 48 B as our parser had it.
- For single-segment models (mnist/SSD/YOLO/segnet) the bug was silent because the kernel Forward path reads `astSrc`/`astDst` from the user's blobs, never from `pstModel->astSeg[i].astDstNode[]`. mnist + SSD remain byte-identical to vendor on av300 after this fix.
- For two-segment models (pvanet, R-FCN, alexnet-frcnn, fasterrcnn_double_roipooling — every two-stage detector), `seg[1]` landed in the middle of `seg[0]`'s src-node body and decoded garbage. After the fix it matches vendor exactly:

  ```
  before: seg[1]: netType=224 srcNum=0 dstNum=0 inst@0x3       len=0x10000
  after:  seg[1]: netType=1   srcNum=1 dstNum=2 inst@0x14b7200 len=0x96f8
  ```

## What end-users get
Anyone calling `HI_MPI_SVP_NNIE_LoadModel` on a two-stage detector model now gets correctly-decoded `astSeg[1]` metadata: real netType, real src/dst counts, real node names, real instruction offsets. Before, those fields contained random bytes from the middle of `seg[0]`'s "data" node payload. Apps that pre-allocated based on these fields would have over- or under-allocated unpredictably.

Forward-dispatching single-shot models is unchanged (no kernel-path regression; verified on av300 against vendor `sample_nnie_main` for mnist + SSD — both byte-identical to vendor baseline).

## How
Walks segments sequentially: read 16 B seg record at `p`, then iterate `src_num + dst_num` nodes advancing by `NNIE_WK_NODE_STRIDE = 64` bytes per node. Next segment record begins where the previous segment's last separator ended.

## Test plan
- [x] Host unit test (`test_loadmodel`): synth `.wk` updated to match real vendor layout; new multi-segment fixture asserts `seg[1]` decode (`netType=1`, src/dst counts, node names).
- [x] Host unit test (`test_get_tskbuf_size`): unaffected — uses `astSrc` info which we already parsed correctly.
- [x] On-target (av300, `test_get_tskbuf_size_on_target` against real pvanet `.wk`):
  ```
  seg[0]: netType=0 srcNum=1 dstNum=4 inst@0x13e2000 len=0xd51e0
  seg[1]: netType=1 srcNum=1 dstNum=2 inst@0x14b7200 len=0x96f8
  GetTskBufSize: seg[0]=96 B, seg[1]=10624 B
  ```
- [x] On-target regression: mnist still classifies `0:4096`; SSD still emits class 7 + class 12 detections at the same coords / confidences as vendor.

## Out of scope
- The kernel-side `ForwardWithBbox` handler (#26) still returns `-EOPNOTSUPP`; this PR just makes the model metadata available to userspace correctly.

### DIFF
--- a/libraries/nnie_neo/include/nnie_wk_format.h
+++ b/libraries/nnie_neo/include/nnie_wk_format.h
@@ -76,14 +76,25 @@ _Static_assert(sizeof(nnie_wk_seg_record_t) == 16,
 	       "nnie_wk_seg_record_t must be exactly 16 bytes");
 
 /* Per-node record (48 bytes), one per src then one per dst, immediately
- * after the segment record(s):
- *   [+0..+11]  u32 height, u32 width, u32 chn
- *   [+12..+13] padding
- *   [+14]      u8  enType (SVP_BLOB_TYPE_E)
- *   [+15]      u8  node id
- *   [+16..+47] szName[32]
+ * after the segment record. Each node body is followed by a 16-byte
+ * inter-node separator on disk:
+ *   [+0..+47]  body
+ *     [+0..+11]  u32 height, u32 width, u32 chn
+ *     [+12..+13] padding
+ *     [+14]      u8  enType (SVP_BLOB_TYPE_E)
+ *     [+15]      u8  node id
+ *     [+16..+47] szName[32]
+ *   [+48..+63] separator (mostly zeros; first byte sometimes carries a
+ *                         node-link count, but vendor's runtime parser
+ *                         doesn't read it)
+ *
+ * For a segment with N total nodes, the on-disk footprint after the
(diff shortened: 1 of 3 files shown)


widgetii:
Closing — behaviour-neutral. Apps using LoadModel for single-shot inference are unaffected, and apps using it for two-stage detectors can't actually run them anyway (kernel bbox path is -EOPNOTSUPP). 'Metadata fields decode correctly' isn't user value if no caller uses those fields today. Will re-open as part of the kernel bbox impl PR if that ever lands.

Issue #183, opened 2026-05-23
cv200/V2A: sensor i2c called from inside ISP_ISR — intermittent boot brick on hi3518ev200


widgetii:
## Symptom

On hi3518ev200 (V2A, `BR2_OPENIPC_SOC_FAMILY=hi3516cv200`) running a fresh OpenIPC firmware, majestic intermittently fails to produce JPEG frames after boot. `curl http://localhost/image.jpg` returns **HTTP 000 / 10 s timeout** on a fraction of boots, while the kernel itself is up, SSH works, `/api/v1/config` returns the web UI, and `pidof majestic` shows it running. `/proc/interrupts` on a bricked boot shows only `VPSS0` ticking — `VI_*` and `MIPI_RX` IRQs are absent, meaning the sensor isn't producing frames at all.

dmesg always carries this WARN (one-shot, at startup):

```
WARNING: CPU: 0 PID: <majestic> at kernel/locking/rtmutex.c:1545 rt_mutex_trylock+0x4c/0xe4
[<c03904d8>] (rt_mutex_trylock) from [<c025cbb8>] (i2c_transfer+0x50/0xac)
[<c025cbb8>] (i2c_transfer) from [<c025cc44>] (i2c_master_send+0x30/0x40)
[<c025cc44>] (i2c_master_send) from [<bf3cf0a0>] (hi_sensor_i2c_write+0xa0/0x110 [open_sensor_i2c])
[<bf3cf0a0>] (hi_sensor_i2c_write) from [<bf242dc4>] (ISP_DRV_RegConfigSensor+0x12c/0x2ac [open_isp])
[<bf242dc4>] (ISP_DRV_RegConfigSensor) from [<bf24332c>] (ISP_IntBottomHalf+0x3e8/0x430 [open_isp])
[<bf24332c>] (ISP_IntBottomHalf) from [<bf2400a8>] (ISP_ISR+0x98/0xd4 [open_isp])
[<bf2400a8>] (ISP_ISR) from [<c003b2d4>] (__handle_irq_event_percpu+0x3c/0x118)
```

## Root cause

[...]

PR #196, opened 2026-05-27
kernel/isp/hi3516cv200: move ISP_IntBottomHalf to a threaded IRQ (V2A i2c-in-IRQ fix)

widgetii:
## Summary

V2A (`hi3518ev200`) boards have always called `ISP_IntBottomHalf` from `ISP_ISR` synchronously, which routes `ISP_DRV_RegConfigSensor → hi_sensor_i2c_write → i2c_transfer → rt_mutex_trylock` through IRQ context. `rt_mutex_trylock` (kernel/locking/rtmutex.c:1545) does:

```c
if (WARN_ON_ONCE(in_irq() || in_nmi() || in_serving_softirq()))
    return 0;
```

so the trylock returns 0 and `i2c_transfer` returns `-EAGAIN`. Sensor exposure/gain writes silently fail. Each sensor reacts differently:

- **OV2735** (issue OpenIPC/firmware#2144 reporter): sensor's internal AGC winds up to max gain → image >96 % white.
- **soi_f22** (lab cam): sensor stays at boot-default exposure → image dim, but `/image.jpg` burst is "stable" only because nothing is actually changing.

The historical `int_bottomhalf` module parameter offered a tasklet path (`int_bottomhalf=1`), but tasklets run in softirq, and `WARN_ON_ONCE(... || in_serving_softirq())` fires there too. Neither path was safe.

`schedule_work()` avoids the WARN but `system_wq` is SCHED_NORMAL (competes with userspace) and `work_struct` coalesces: multiple ISR-schedules while a work is queued/running collapse into a single bottom-half run, with per-IRQ `pstDrvCtx->stIntSch` getting overwritten. The `_GPL` workqueue helpers (`system_highpri_wq`, `cancel_work_sync`, `flush_work`, `alloc_workqueue`) are unavailable to MPP_LICENSE.

## Fix

Move `ISP_IntBottomHalf` into a threaded IRQ via `request_threaded_irq()`:

- `ISP_ISR` returns `IRQ_WAKE_THREAD` unconditionally.
- A dedicated kthread (`irq/<vec>-ISP`, SCHED_FIFO prio 50) runs the bottom half in process context — the only legal caller for `rt_mutex_trylock`.
- Threaded IRQs do **not** coalesce: the kernel queues the thread wake per hardirq, so per-frame sensor writes land per frame at ~30 Hz.
- `request_threaded_irq` is `EXPORT_SYMBOL` (kernel/irq/manage.c) — callable from MPP_LICENSE.
- The `int_bottomhalf` module parameter is preserved for ABI compat (`/sys/module/open_isp/parameters/int_bottomhalf`) but is now a no-op; default flipped to `HI_TRUE` to reflect the new semantics.

## Verification on `openipc-hi3518ev200.dlab.torturelabs.com`

Sensor: `soi_f22`. Pulled the patched `hi3518e_isp.ko` onto the cam via overlay and rebooted between each variant; same scene throughout.

| Variant | `rt_mutex_trylock` WARN | JPEG burst variance | image |
|---|---|---|---|
| rom #186 unmodified (tasklet, `int_bottomhalf=0`) | 1× (`WARN_ON_ONCE`) | ±0.2 % (52.9–53.2 KB) | dim, sensor stuck |
| `schedule_work` (intermediate) | 0 | ±10 % (105–116 KB) | working but coalesces, AE step changes |
| **`request_threaded_irq` (this PR)** | **0** | **±2 % (113–116 KB)** | working, smooth, `irq/38-ISP` visible in `ps` |

`/dev/openipc-frame-ts` still emits 32-byte events with monotonic seq at frame rate — the `g_stIspFtTasklet` introduced by #186 is untouched and stays in softirq (its body is a pure `spinlock_irqsave` + ring write, no rt_mutex).

The minor residual AE oscillation visible on soi_f22 after this fix is a vendor AE-tuning artifact, not a kernel bug — sensor writes are now actually landing so the AE control loop is closed (it was open before, with writes failing silently). This is the same effect that should bring the OV2735 image back from "all white" to proper exposure in the reporter's setup.

## Closes

- OpenIPC/firmware#2144

## Refs

#155, #178, #182, #183, #186 — the chain that exposed the latent V2A i2c-in-IRQ race.

### DIFF
--- a/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
+++ b/kernel/isp/arch/hi3516cv200/firmware/drv/isp.c
@@ -112,6 +112,7 @@ extern HI_BOOL VB_IsSupplementSupport(HI_U32 u32Mask);
 
 static HI_U32 g_u32ISP_STAT_DRC_PHY = HI_NULL;
 void ISP_IntBottomHalf(unsigned long data);
+static irqreturn_t ISP_IntBottomHalf_threaded(int irq, void *dev_id);
 
 int isp_irq = -1;
 
@@ -129,7 +130,7 @@ void __iomem            *reg_isp_base_va = HI_NULL;
 HI_U32                  proc_param = 0;        /* 0: close proc; n: write proc info's interval int num */
 HI_U32                  pwm_num = 1;
 HI_U32                  update_pos = 0;         /* 0: frame start; 1: frame end */
-bool                    int_bottomhalf = HI_FALSE;  /* 1 to enable interrupt processing at bottom half */
+bool                    int_bottomhalf = HI_TRUE;   /* preserved for ABI; bottom half is always deferred to a threaded IRQ now */
 HI_U32                  lsc_update_mode = 0;
 
 spinlock_t g_stIspLock;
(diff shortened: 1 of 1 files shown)


widgetii:
Closing in favour of #197.

#196 (threaded-IRQ deferral of `ISP_IntBottomHalf`) was treating a symptom — it made sensor writes succeed by routing them through process context, where `rt_mutex_trylock` doesn't WARN. But that was only necessary because something else had silently broken the synchronous IRQ path.

The real cause is in `kernel/sensor_i2c/hi3516cv200/sensor_i2c.c`: `abadb7cf` (#162) added a `#ifndef hi_i2c_master_send` macro check around a `#define hi_i2c_master_send i2c_master_send` fallback. `hi_i2c_master_send` is exported by the vendor's `drivers/i2c/busses/i2c-hibvt.c` as an `extern` function, not a preprocessor macro, so the `#ifndef` always evaluated to true and the alias fired on vendor kernel too. The vendor function bypasses i2c-core locking (calls `hibvt_i2c_xfer` directly), the alias forced everything through `i2c_transfer → rt_mutex_trylock` and broke from IRQ.

With #197 the vendor extern is used again, `ISP_DRV_RegConfigSensor → hi_sensor_i2c_write → hi_i2c_master_send` is IRQ-safe like the vendor designed it, the synchronous bottom half works, and the rt_mutex WARN goes away without any deferral. Verified on the same lab `openipc-hi3518ev200` bench: WARN count 3 → 0, frame brightness stdev 0.16 over 200 frames.

PR #201, opened 2026-05-28
openipc_frame_ts: tunable ring_depth module_param (closes #200)

### LINKED ISSUE #200: cv500/hi3516av300 frame-ts chrdev: zero MIPI_RX IRQs, zero coalesced events, ring drops under reader load
Surfaced while verifying #155 from the majestic side (widgetii/majestic#83 / widgetii/majestic#86). The SR-anchor functionality of #155 works on cv500 — drift signature matches the hi3516ev300 reference. But the chrdev counters and `/proc/interrupts` numbers on this board don't match what #155's design assumed.

## Setup

- Board: `openipc-hi3516av300.dlab.torturelabs.com` (cv500 family)
- Sensor: imx415 over MIPI, 20 fps, h264
- Kernel: `4.9.37 #1 SMP Wed May 27 23:56:21 UTC 2026 armv7l`
- Modules loaded: `open_openipc_frame_ts` (from #155 merged), `open_isp`, full HiSilicon MPP stack
- chrdev: `/dev/openipc-frame-ts` minor=58, event types=2 (MIPI_FS + ISP_FEND mask supported)
- majestic: master+f5207b6, with the `openipc_frame_ts_consumer.c` from widgetii/majestic#87 (v2 ABI). Consumer fd persistently open, "SR will anchor to sensor frame-start..." marker prints.

## Observation 1 — `/proc/interrupts MIPI_RX` stays at 0 since boot

```
 43:          0          0     GIC-0  89 Level     MIPI_RX
 45:      81374          0     GIC-0  88 Level     VI_CAP0
 46:      40681          0     GIC-0  76 Level     VI_PROC0
```

Sampled 30 s apart with `S95majestic` running and a live RTSP session: MIPI_RX delta = 0; VI_CAP0 increments ~40 Hz (~2× framerate); VI_PROC0 ~20 Hz (= framerate). Yet the chrdev consumer is clearly receiving MIPI_FS events — `dropped` accumulates under load (see Observation 3), meaning the kernel is calling `openipc_frame_ts_push(.., OPENIPC_FT_EVT_MIPI_FS)` from somewhere.

Contrast with the same board's reading in #177:
```
MIPI_RX    IRQ 43: 64371   ← per-frame (matches sensor rate)
```

Either the cv500 MIPI_RX hook moved off IRQ 43 between then and the current build, or the per-SoC FS push is now sourced from a different IRQ path (VI_CAP0 candidate, ~2× framerate would explain the high push rate that drives Observation 3). Worth nailing down which IRQ the cv500 build is actually using for `OPENIPC_FT_EVT_MIPI_FS`.

## Observation 2 — `OPENIPC_FT_IOC_GET_COALESCED` stays at 0

The cv500 dedupe rationale in #155 cites ~4% of frames double-firing the MIPI vsync IRQ at ~30-80 µs intervals, absorbed by the 1 ms per-channel dedupe window. Over a full verification run (multiple RTSP sessions, two 30 s correlator passes, ~30 min uptime), `GET_COALESCED` returned **0**.

If Observation 1's hypothesis is correct (FS events sourced from VI_CAP0 not MIPI_RX on this build), the doubles described in #155 may be specific to the MIPI_RX path and simply don't manifest from the alternative source. Functionally fine — fewer doubles to dedupe means no harm — but the design assumption "cv500 needs ~4% dedupe" doesn't hold on this build, which is worth knowing for sizing decisions.

(Sanity: `GET_COALESCED` ioctl itself returns success; the counter is just 0. So it's not an ioctl wiring bug.)

## Observation 3 — `GET_DROPPED` accumulates under reader load

Counter readings:
- Idle (no RTSP clients, majestic libevent base running): **0 drops/s** over a 30 s window — chrdev drains cleanly.
- During the active correlator runs (two 30 s RTSP sessions back-to-back): `dropped` went from 3758 → 7403 = **+3645 drops over ~70 s of busy event loop**, i.e. ~52 drops/s.

The consumer's `on_readable` drains in batches of 8 per dispatch (`for (;;) read(fd, buf[8], ...)` until EAGAIN). It's not under-draining; it's not dispatched often enough. Under RTSP+RTP+RTCP traffic the libevent base spends more time per dispatch and the kernel ring (which appears small) overflows in between.

For #83's SR-anchor use case this is **non-impacting** — the consumer cache only holds the latest event per channel and is read every ~5 s at SR-tick time. With the 100 ms `max_age` staleness gate, drops at this rate still leave a fresh enough cache. The empirical SR drift was +73 ms median, +68 to +88 ms range — no fallback to `clock_gettime()` observed, which would have shown up as drift jumps to the unanchored (~−60 ms) regime.

But for any future consumer that wants every frame's `(pts_us, wall_ns)` — per-frame evidence chains, frame-by-frame latency histograms, multi-sensor sync correlators — the ring size as currently configured won't suffice on cv500 under load. Worth sizing the ring (or making it sizable via module param) before such consumers ship.

## Reproduction

```sh
# On the cv500 board:
lsmod | grep open_openipc_frame_ts                  # module present
ls -la /dev/openipc-frame-ts                         # chrdev present
cat /proc/interrupts | grep -E "MIPI_RX|VI_CAP0|VI_PROC0"

# Cross-compile tiny ioctl helper (host side, arm-himix200-linux-gcc -static):
#   ioctl(fd, _IOR('o',2,u64), &dropped)
#   ioctl(fd, _IOR('o',4,u64), &coalesced)
# Push, run before/after a busy majestic RTSP session.

# Hold an RTSP session ~30s with majestic streaming, then snapshot counters.
```

## Suggested follow-ups (no fix proposed here, just framing)

1. Confirm which IRQ source cv500 currently routes `OPENIPC_FT_EVT_MIPI_FS` through (instrument `openipc_frame_ts_push` at module load to log the IRQ it's hooked off of, or just grep the per-SoC hook code).
2. If FS is now sourced from VI_CAP0 (~2× framerate), is one of those two pushes the actual sensor-FS-equivalent and the other something else (line-start? half-frame? skip?) that should be filtered before push? If so, the high apparent push rate driving Observation 3's drops would resolve.
3. Ring size on the chrdev: currently sufficient for the SR-anchor consumer pattern; document a `module_param` knob or auto-size based on the push rate observed at init.

The merged #155 + widgetii/majestic#83 chain is working as designed for the SR-anchor use case on cv500 — this issue is to track the divergence between the design assumptions and what the running cv500 build actually does, so future consumers and other SoC families can be sized correctly.

xref: #155 (chrdev introduction + dedupe), #177 (FEND IRQ routing on the same family), widgetii/majestic#83 (SR-anchor consumer), widgetii/majestic#86 (cv500 verification — closed).

widgetii:
## Summary

Per-channel chrdev ring depth was a compile-time `#define OPENIPC_FT_DEPTH 256`. Sufficient for the SR-anchor consumer in #155 (reads cached state every ~5 s, tolerates stale entries). Too small for per-frame consumers — frame-by-frame latency histograms, evidence chains, multi-sensor sync correlators — under bursty reader-side dispatch.

Empirical on cv500 / hi3516av300_lite (issue #200 Obs 3):

| ring_depth | reader pause | drops/s |
|---|---|---|
| 256 (current) | 8 s | 8.1 |
| 256 (current) | 15 s | 23.0 |
| 4096 (new param) | 8 s | 0.0 |
| 4096 (new param) | 15 s | 0.0 |
| 4096 (new param) | 60 s | 0.0 |

Lab cam: `openipc-hi3516av300.dlab.torturelabs.com`, imx415 20 fps, push rate ~40/s after #199 moved MIPI_FS to the ISP_INT_FE_FSTART edge-detect (20 FSTART + 20 FEND).

## What changed

- `OPENIPC_FT_DEPTH` (compile-time) → `ring_depth` module_param (default 256, range [16, 65536], rounded up to next power-of-2 at `__init` for the head/tail wrap mask).
- Per-channel rings move from inline static array to `kmalloc_array` in `__init`, freed in `__exit`. Init unwinds on partial-failure paths.
- `ring_count` / `ring_full` / `pop_one` / `openipc_frame_ts_push` use `g_ring_mask` instead of `(OPENIPC_FT_DEPTH - 1)`.
- `dmesg` prints the final depth at module init: `openipc_frame_ts: /dev/openipc-frame-ts minor=58, event types=2, ring_depth=4096`.
- Sysfs exposes the value at `/sys/module/open_openipc_frame_ts/parameters/ring_depth` (0444).

## Non-disruptive at default params

Default ring_depth=256 keeps the binary behaviour identical to before: same depth, same wrap math, same drop semantics. Only the storage moved from BSS to heap. Memory footprint unchanged for unmodified deployments (8 KiB/chn × 8 chn = 64 KiB).

## Operator workflow

```
# at module insert
insmod open_openipc_frame_ts.ko ring_depth=4096

# or persistent
echo 'options open_openipc_frame_ts ring_depth=4096' > /etc/modprobe.d/openipc_frame_ts.conf
```

depth=4096 ≈ 8 s buffer at 240 fps × 2 types; 128 KiB/chn × 8 chn = 1 MiB worst case.

## Verified on lab cam

- `cat /sys/module/open_openipc_frame_ts/parameters/ring_depth` → `4096`
- `dmesg | grep openipc_frame_ts` → `ring_depth=4096`
- Lazy-reader (drains in 64-event batches with configurable pause) at 8 s, 15 s, 60 s pauses: 0 drops in all three cases at depth=4096; reproduces 8.1 / 23.0 drops/s at depth=256.

## Test plan

- [x] Builds clean on cv200 (hi3516cv200_lite, kernel 4.9.37, ARMv6 musl).
- [x] Builds clean on cv500 (hi3516av300_lite, kernel 4.9.37 SMP, ARMv7 musl p2v8).
- [x] Module loads on cv500 with default ring_depth=256, behaves identically.
- [x] Module loads on cv500 with ring_depth=4096, dmesg prints the depth.
- [x] sysfs param exposed at `/sys/module/open_openipc_frame_ts/parameters/ring_depth`.
- [x] At depth=4096, 15 s reader-pause yields 0 drops where depth=256 yields 23 drops/s.

xref: #155 (chrdev introduction), #199 (cv500 / V4 MIPI_FS push moved to ISP_INT_FE_FSTART — the non-disruptive event source driving the push rate that #200 measured).

### DIFF
--- a/kernel/openipc_frame_ts/openipc_frame_ts.c
+++ b/kernel/openipc_frame_ts/openipc_frame_ts.c
@@ -39,12 +39,33 @@
 #define OPENIPC_FT_MAX_CHN        8
 #define OPENIPC_FT_MAX_EVT_TYPE   2   /* MIPI_FS + ISP_FEND so far */
 /*
- * Per-channel ring depth (power of 2). 256 events ≈ 0.5 s of buffer at
- * 240 fps × 2 event types — covers high-fps modes (imx335 800x480 240 fps)
- * with margin for reader-side jitter (scheduler latency, printf-to-disk).
- * At 32 B/event the per-channel cost is 8 KiB.
+ * Per-channel ring depth bounds. Must be powers of two — push uses
+ * `head & (depth-1)` wrap math, which only works for power-of-2 depths.
+ *
+ * Default 256 ≈ 0.5 s buffer at 240 fps × 2 event types (8 KiB/chn). Fine
+ * for the SR-anchor use case (reads cached state every ~5 s and tolerates
+ * stale entries).
+ *
+ * For per-frame consumers — frame-by-frame latency histograms, evidence
+ * chains, multi-sensor sync correlators — the default may be undersized
+ * if the reader's dispatch latency is bursty (libevent under RTSP+RTCP
+ * load on cv500 was observed to lose ~52 events/s during 30 s sessions,
+ * see openhisilicon#200). Raise via `insmod open_openipc_frame_ts
+ * ring_depth=4096` (16× default, ~8 s buffer; 128 KiB/chn × 8 chn = 1 MiB
+ * total worst case).
  */
-#define OPENIPC_FT_DEPTH          256
+#define OPENIPC_FT_DEPTH_MIN      16
+#define OPENIPC_FT_DEPTH_MAX      65536
+#define OPENIPC_FT_DEPTH_DEFAULT  256
+
+static unsigned int ring_depth = OPENIPC_FT_DEPTH_DEFAULT;
+module_param(ring_depth, uint, 0444);
+MODULE_PARM_DESC(ring_depth,
+	"Per-channel ring depth (power of 2, 16..65536, default 256). "
+	"Bump for per-frame consumers under bursty reader load.");
+
+static unsigned int g_ring_depth;
+static unsigned int g_ring_mask;
 /*
  * Drop double-pushes of the SAME event type within these intervals.
  *
@@ -74,7 +95,7 @@ static const u64 openipc_ft_min_interval_ns[OPENIPC_FT_MAX_EVT_TYPE] = {
 };
 
 struct openipc_ft_chn {
-	struct openipc_frame_ts_event ring[OPENIPC_FT_DEPTH];
+	struct openipc_frame_ts_event *ring;   /* kmalloc'd, g_ring_depth entries */
 	unsigned int head;
 	unsigned int tail;
 	u32          seq[OPENIPC_FT_MAX_EVT_TYPE];
@@ -170,12 +191,12 @@ static void strobe_on_event(unsigned int event_type)
 
 static inline unsigned int ring_count(const struct openipc_ft_chn *c)
 {
-	return (c->head - c->tail) & (OPENIPC_FT_DEPTH - 1);
+	return (c->head - c->tail) & g_ring_mask;
 }
 
 static inline bool ring_full(const struct openipc_ft_chn *c)
 {
-	return ring_count(c) == (OPENIPC_FT_DEPTH - 1);
+	return ring_count(c) == g_ring_mask;
 }
 
 static inline bool ring_empty(const struct openipc_ft_chn *c)
@@ -216,7 +237,7 @@ void openipc_frame_ts_push(unsigned int vi_chn, unsigned int event_type)
 
 	if (ring_full(c)) {
 		/* drop oldest */
-		c->tail = (c->tail + 1) & (OPENIPC_FT_DEPTH - 1);
+		c->tail = (c->tail + 1) & g_ring_mask;
 		c->dropped++;
 	}
 	e = &c->ring[c->head];
@@ -226,7 +247,7 @@ void openipc_frame_ts_push(unsigned int vi_chn, unsigned int event_type)
 	e->seq        = c->seq[event_type]++;
 	e->event_type = event_type;
 	e->reserved   = 0;
-	c->head = (c->head + 1) & (OPENIPC_FT_DEPTH - 1);
+	c->head = (c->head + 1) & g_ring_mask;
 	spin_unlock_irqrestore(&c->lock, flags);
 
 	strobe_on_event(event_type);
@@ -312,7 +333,7 @@ static bool pop_one(u32 chn_mask, u32 evt_mask,
 			struct openipc_frame_ts_event *cand =
 				&c->ring[c->tail];
 
-			c->tail = (c->tail + 1) & (OPENIPC_FT_DEPTH - 1);
+			c->tail = (c->tail + 1) & g_ring_mask;
 			if (event_passes(cand, chn_mask, evt_mask)) {
 				*out = *cand;
 				got = true;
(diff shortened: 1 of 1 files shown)


widgetii:
Retested on `openipc-hi3516av300.dlab.torturelabs.com` (imx415 @ 20 fps, IRQ source = ISP_INT_FE_FSTART per #199, push rate ~40 Hz) with this PR's modules live.

State confirmation:
```
$ cat /sys/module/open_openipc_frame_ts/parameters/ring_depth
4096
$ dmesg | grep openipc_frame_ts
openipc_frame_ts: /dev/openipc-frame-ts minor=58, event types=2, ring_depth=4096
```

The synthetic test in the PR description used a lazy 64-batch reader; majestic's real-world consumer (`openipc_frame_ts_consumer.c` from widgetii/majestic#87) drains in 8-event batches from libevent `EV_READ`. Wanted to confirm the fix holds for the smaller batch size under actual RTSP load.

Method: 20 s ffmpeg warmup to drain the boot backlog and reach steady-state, then bracketed a 30 s `manual_rtcp_correlator.py` run with `GET_DROPPED` + `GET_COALESCED` ioctl snapshots and `/proc/interrupts` snapshots either side.

| Metric | pre-#201 (depth=256) | post-#201 (depth=4096) |
| --- | --- | --- |
| `dropped` Δ over 30 s busy event loop | +3645 (~52/s) | **+0** |
| `coalesced` Δ | 0 | 0 |
| VI_CAP0 IRQ rate | ~40 Hz | ~41.2 Hz |
| VI_PROC0 IRQ rate | ~20 Hz | ~20.6 Hz |
| MIPI_RX IRQ rate | 0 | 0 |
| SR drift median (NTP-corrected) | +73.0 ms | +84.3 ms |
| SR drift p95 (NTP-corrected) | +85.5 ms | +96.7 ms |

Drops fully resolved at the 8-batch reader size, exactly as the synthetic test predicted. SR-anchor drift unchanged in shape — the +11 ms shift in median is run-to-run NTP discipline variance on the host workstation, not signal.

LGTM for the cv500 / hi3516av300 path. Closes the Obs 3 follow-up from widgetii/majestic#86 cleanly.

PR #209, opened 2026-08-15
acodec: retune the ADC when a clock change invalidates it

### LINKED ISSUE #208: acodec (v750): kernel module skips the staged analog power-up U-Boot does, so the codec calibrates against an unsettled vref
## Summary

The kernel `acodec` module for the v750 inner codec (gk7205v200 / hi3516ev200 family) powers the entire analog front end up in a single register write, where the vendor's own U-Boot build of the *same* driver uses a staged sequence with a 20 ms vref settle. The consequence is measurable on real hardware: the codec's RC calibration latches a bad tuning code and the ADC sigma-delta modulator settles into a limit cycle — an audible tone at `fs·39/1024` plus harmonics.

Surfaced while investigating OpenIPC/majestic#285, where it is a ~+20 dB comb on the mic on every cold boot.

## Evidence

`kernel/acodec/gk7205v200/acodec_log.o` is a prebuilt vendor object but keeps its symbols, so `acodec_init()` disassembles cleanly. It does:

1. ioremap `0x100F0000`
2. write the powered-down analog defaults — `ANAREG0 = 0x3434DFFF` (every `pd_*` bit set except `pd_ldo`), `ANAREG1 = 0xDF605E65`, `ANAREG2 = 0x00255548`, `ANAREG3 = 0x04293B50` (bit0 `rstn` = 0), plus `DIGCTRL1..5`
3. `osal_udelay(1000)` — **1 ms**
4. `acodec_soft_reset()`, which writes `ANAREG0 = 0x1C1C0000` — every `pd_*` bit cleared, so vref, ibias, LDO, micbias, linein and both ADCs come up **in one store** — and `ANAREG3 = 0x072D0451` (`rstn` = 1)

Compare `acodec_device_init()` in the vendor U-Boot source for the same v750 block (`product/audio/acodec/v750/acodec.c` in the Goke SDK), which calls `acodec_ana_power_on()`:

```
pd_ldo/pd_ibias/pd_ctcm/pd_dac_vref  -> 0
pd_vref -> 0 ; vref_pu_pdb -> 1
udelay(20 * 1000)                      <-- 20 ms vref settle
pop_rst -> 0
pd_lineoutl/r -> 0
udelay(10 * 1000)
pop_lineout_pull_en -> 0
pd_micbias1/2 -> 0
pd_rctune -> 0 ; en_rctune -> 1
udelay(1 * 1000)
pd_linein_l/r, pd_adcl/r -> 0
rstb_dac -> 1
```

The kernel module runs none of this. It also means the RC calibration performed at insmod happens against a vref that has had 1 ms rather than 20 ms, and against the AIAO clock as it stands before any `HI_MPI_AI_Enable` has programmed it.

## Measured impact

Lab gk7205v200 and hi3516ev300, 16 kHz mono, raw PCM off the camera, first streamer start after a cold boot (warm reboot and hard PoE power cycle both):

- coherent tone at 609.375 Hz = `fs·39/1024` with five harmonics, up to +20 dB over the noise floor
- calibration result readback at `0x100F0018` reads `0x23`/`0x25` (bit 3 clear) in that state, `0x2B`/`0x2F` (bit 3 set) once calibration is redone correctly
- forcing a recalibration by toggling `ANAREG1` bit 1 (`en_rctune`) with the clock live takes the capture from 36 LSB rms with the comb to 6 LSB with no coherent tone

Full data in OpenIPC/majestic#285.

## What could be done here

The staged power-up lives inside the prebuilt vendor object, so this is not a one-line wrapper change:

1. **Rebuild the acodec driver from source** and port `acodec_ana_power_on()` from the U-Boot v750 sources. This is the actual fix — the codec should not be calibrating against an unsettled vref.
2. Failing that, **document the constraint**: anything using the inner codec must retrigger `ACODEC_SET_I2S1_FS` after `HI_MPI_AI_Enable`, because the calibration done before that point is against the wrong clock.

The consumer side has been worked around already, so nothing here is urgent — but as long as the driver stands, every consumer of the inner codec has to know to recalibrate.

Chips affected: whatever ships this v750 acodec — confirmed on gk7205v200 and hi3516ev200/ev300.

widgetii:
Closes #208 (the mechanism, not the framing — see below).

## What happens

The v750 inner codec tunes its ADC RC network whenever `ACODEC_SET_I2S1_FS` pulses `en_rctune`, and the code it latches is only correct for the codec clock that was running at the time. Register `0x18` reports that live: bit 3 stays set while the tuning still matches the clock. Nothing in the driver reads or writes that offset — I disassembled `acodec_log.o` to check — so it is pure hardware status.

Enabling an AI device reprograms the AIAO TX0 clock, which invalidates a tuning that was correct a moment earlier. Traced on gk7205v200 by sampling the status and clock registers continuously across a cold boot; three distinct states in the entire trace:

```
t=39.39  0x18=0x3F  AIAO140=0x000DFB24 AIAO144=0x00000131  idle
t=43.31  0x18=0x3F  AIAO140=0x000DFB24 AIAO144=0x00000131  codec configured
t=43.38  0x18=0x25  AIAO140=0x00095218 AIAO144=0x0000C117  AI enabled
```

The calibration never fails — it is valid at t=43.31 and invalidated 70 ms later. Nothing recalibrates afterwards, so the ADC modulator runs mistuned and settles into a limit cycle: an audible tone at `fs*39/1024` and its harmonics, up to 20 dB over the noise floor, for as long as the device stays up. Reported downstream as OpenIPC/majestic#285.

**This corrects #208 as I first filed it.** I blamed the missing staged vref settle. That cannot be the cause: the module loads tens of seconds before any consumer starts, so vref has long settled by the time the invalidating calibration happens — and the original reporter had already shown that delaying the consumer's first start by 40 s changes nothing. The staged-power-up deviation from the vendor's U-Boot build of the same driver is real and worth fixing on its own, but it is a separate item.

## The change

The codec cannot observe the clock moving, but it can observe the result. Poll the validity bit on a 1 s timer and re-pulse `en_rctune` when it goes stale — the same pulse `SET_I2S1_FS` issues. The tuning then follows the clock whichever consumer moves it, or when.

Backs off to 10 s after 10 consecutive failures rather than pulsing every second over a codec somebody may be reading from, logs the first retune and the back-off, and `rctune_watchdog=0` disables it. Guarded to the parts whose status register is known — `gk7205v200` and `hi3516ev200`, both confirmed on hardware; `ccflags-y` already defines `-D$(CHIPARCH)`, so the chip name is available as a macro. It maps the codec block itself rather than depending on `pAcodec`, which `open_aio` exports but may leave NULL.

## Verification

gk7205v200, 16 kHz mono, first consumer start after a cold boot. Capture rms and the 609.375 Hz fundamental over the local noise floor:

| acodec module | consumer | `0x100F0018` | rms (LSB) | 609 Hz | driver log |
|---|---|---|---|---|---|
| stock | unpatched | `0x25` | 41.4 | +3.8 dB, harmonics to +11.7 | — |
| this | unpatched | `0x2E` | 5.2 | +3.2 dB, no comb | one retune, logged |
| this | retunes itself | `0x2E` | 5.4 | +3.5 dB, no comb | silent over 75 s |

Row 2 is the point: an **unpatched** consumer, comb gone by itself, with

```
acodec: ADC tuning stale after a clock change, retuning
```

The row 1 control **reloads the stock module** immediately before the consumer starts and still comes up mistuned. That control matters: the blob's own init writes `en_rctune` 0 then 1 a millisecond apart, which is a real pulse, so without it a reload could have been what cured things. It is not — the invalidating clock change happens afterwards.

Row 3 is a consumer already fixed to retune after enabling AI. The watchdog costs two register reads a second and never fires; the two fixes are independent and do not fight.

Built against openipc/linux `goke-gk7205v200` (4.9.37), `CHIPARCH=gk7205v200 KO_LOG=nolog`; vermagic matches the shipped module exactly.

### DIFF
--- a/kernel/acodec/acodec_init.c
+++ b/kernel/acodec/acodec_init.c
@@ -3,18 +3,158 @@
 #include <linux/printk.h>
 
 #include "type.h"
+#include "osal.h"
 
 extern int acodec_init(void);
 extern void acodec_exit(void);
 extern unsigned int init_delay_time;
 module_param(init_delay_time, uint, S_IRUGO);
 
+#if defined(gk7205v200) || defined(hi3516ev200)
+#define ACODEC_RCTUNE_WATCHDOG
+#endif
+
+#ifdef ACODEC_RCTUNE_WATCHDOG
+
+/*
+ * The inner codec tunes its ADC RC network whenever ACODEC_SET_I2S1_FS pulses
+ * en_rctune, and the code it latches is only correct for the codec clock that
+ * was running at the time. Register 0x18 reports that live: bit 3 stays set
+ * for as long as the tuning still matches the clock.
+ *
+ * Enabling an AI device reprograms the AIAO TX0 clock, which invalidates a
+ * tuning that was correct a moment earlier, and nothing recalibrates
+ * afterwards. Traced on gk7205v200 inside a single 70 ms window: 0x18 read
+ * 0x3F with the codec fully configured, then 0x25 the instant
+ * AIAO_CLK_TX0_CRG/CFG moved. From there the ADC modulator runs mistuned and
+ * settles into a limit cycle -- an audible tone at fs*39/1024 and its
+ * harmonics, up to 20 dB over the noise floor, lasting as long as the device
+ * stays up. Confirmed on gk7205v200 and hi3516ev200. See OpenIPC/majestic#285.
+ *
+ * The codec cannot observe the clock moving, but it can observe the result.
+ * Watch the validity bit and retune when it goes stale: that is the same pulse
+ * the SET_I2S1_FS ioctl issues, so the tuning follows the clock no matter
+ * which consumer moves it, or when.
+ */
+
+#define ACODEC_REGS_BASE 0x100f0000
+#define ACODEC_REGS_SIZE 0x1000
+
+#define ACODEC_ANAREG1 0x04
+#define ACODEC_EN_RCTUNE (1u << 1)
+#define ACODEC_CAL_STATUS 0x18
+#define ACODEC_CAL_VALID (1u << 3)
+
+/* One retune settles it, so a second is a long time to sound wrong. Once it is
+ * clear the tuning will not take, stop pulsing every second over a codec
+ * somebody may be reading from. */
+#define ACODEC_POLL_MS 1000
+#define ACODEC_POLL_STUCK_MS 10000
+#define ACODEC_RETUNE_LIMIT 10
+
+static int rctune_watchdog = 1;
+module_param(rctune_watchdog, int, S_IRUGO);
+MODULE_PARM_DESC(rctune_watchdog,
+		 "retune the ADC when a clock change invalidates it (default 1)");
+
+static void *acodec_regs;
+static osal_timer_t rctune_timer;
+static unsigned int rctune_failures;
+
+#define acodec_readl(off) osal_readl((uintptr_t)acodec_regs + (off))
+#define acodec_writel(v, off) osal_writel(v, (uintptr_t)acodec_regs + (off))
+
+static void acodec_rctune(void)
+{
+	unsigned int anareg1 = acodec_readl(ACODEC_ANAREG1);
+
+	acodec_writel(anareg1 & ~ACODEC_EN_RCTUNE, ACODEC_ANAREG1);
+	osal_udelay(30);
+	acodec_writel(anareg1 | ACODEC_EN_RCTUNE, ACODEC_ANAREG1);
+}
+
+static void acodec_rctune_check(unsigned long data)
+{
+	unsigned long period = ACODEC_POLL_MS;
+
+	if (acodec_readl(ACODEC_CAL_STATUS) & ACODEC_CAL_VALID) {
+		rctune_failures = 0;
+	} else {
+		acodec_rctune();
+
+		rctune_failures++;
+		if (rctune_failures == 1)
+			osal_printk("acodec: ADC tuning stale after a clock "
+				    "change, retuning\n");
+		else if (rctune_failures == ACODEC_RETUNE_LIMIT)
+			osal_printk("acodec: ADC tuning will not latch, "
+				    "backing off\n");
+
+		if (rctune_failures >= ACODEC_RETUNE_LIMIT)
+			period = ACODEC_POLL_STUCK_MS;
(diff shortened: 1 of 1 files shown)


widgetii:
Thanks — the timer teardown race is real, and I have fixed it in daa5fdb.

Confirmed against the source before acting. Both branches of `osal_timer.c` do the same thing:

```c
int osal_del_timer(osal_timer_t *timer)   { ... return del_timer(&ot->tl); }
int osal_timer_destory(osal_timer_t *timer){ del_timer(&ot->tl); kfree(ot); ... }
```

`del_timer()` in both, never `del_timer_sync()`, and destroy `kfree()`s on top. My callback re-arms itself unconditionally, so a teardown landing while it runs could free the timer and unmap the codec registers underneath it, and then the callback would write to both. `load_goke`'s `remove_audio` rmmods acodec on every teardown, so it is reachable in the field rather than theoretical.

## What I changed

Moved the watchdog onto delayed work, cancelled synchronously, with a stopping flag checked before re-arming so a cancel cannot be outlived by the work the callback queues:

```c
rctune_stopping = true;
cancel_delayed_work_sync(&rctune_work);
osal_iounmap(acodec_regs);
```

It also gets the 30 µs rctune settle out of a softirq and into process context, which is a better place for it regardless.

## On the suggested fix

The suggestion was to make `osal_del_timer()`/`osal_timer_destory()` synchronous instead. I deliberately did not: that changes semantics for every user of the API, and `del_timer_sync()` deadlocks if any of them delete a timer from inside its own callback — which I have not audited. It is worth fixing, just not from an acodec PR. Happy to file it separately if you want it tracked.

## Re-verified on gk7205v200

- Function unchanged: unpatched consumer, first start after a cold boot → `0x100F0018` = `0x2F`, capture rms 5.1 LSB, no comb, one logged retune.
- Teardown: 30 rmmod/insmod cycles under live audio, with teardown deliberately landing at four different points of the poll period. No oops, no warnings, module reloads clean every time.
- No false positives: restarting the consumer at a different sample rate moves the AIAO divider (`AIAO144` `0xC117` → `0xC119`) without invalidating the tuning, and the watchdog correctly stays quiet.

I did not try to reproduce the original race before fixing it — the window is a fraction of a 1 s period and I would not trust a negative result. The code argument stands on its own.

PR #211, opened 2026-08-15
osal: wait for the timer callback before freeing the timer

### LINKED ISSUE #210: osal: timer delete is not teardown-safe — del_timer() + kfree() can free a running callback
## Summary

`osal_del_timer()` and `osal_timer_destory()` are both `del_timer()`, never `del_timer_sync()`, and destroy `kfree()`s the timer on top. Any caller whose timer callback re-arms itself — the natural shape here, since `osal_set_timer()` is a one-shot `mod_timer()` — can have the timer freed and its resources unmapped while the callback is still running.

Found while reviewing #209 (thanks @qodo). I avoided the API there rather than change it, but there is at least one shipped in-tree driver with the same bug, and the obvious fix is wrong in a way worth writing down.

## The code

`kernel/osal/linux/kernel/osal_timer.c`, identically in both the `COMPAT_TIMER_SETUP` and the pre-4.15 branch:

```c
int osal_del_timer(osal_timer_t *timer)     { ...; return del_timer(&ot->tl); }
int osal_timer_destory(osal_timer_t *timer) { del_timer(&ot->tl); kfree(ot); timer->timer = NULL; return 0; }
```

`del_timer()` does not wait for a callback already running on another CPU. `osal_timer_destory()` then frees the wrapper the callback is about to touch.

## A live example in-tree

`kernel/rtc/hi3519v101/hi_rtc.c`:

- `temperature_timer.function = temperature_detection` (L999)
- `temperature_detection()` (L753) ends at `timer_again: osal_set_timer(&temperature_timer, 1000 * t_second);` (L803) — **it re-arms itself**
- `rtc_exit()` (L1060) does `osal_del_timer()` then `osal_timer_destory()` (L1062-1063)

So on module unload, if `temperature_detection` is mid-flight, `kfree(ot)` happens underneath it and the callback then calls `osal_set_timer()` on `timer->timer`, which is freed and NULLed. Use-after-free plus a NULL deref, on every teardown that loses the race.

## Why "just use del_timer_sync()" is wrong here

That was the suggested fix on #209 and I did not take it, because auditing the callers shows it breaks things:

`kernel/ir/hi3519v101/hiir.c`:

- `repkey_timeout_timer.function = repkey_timeout_handler` (L584)
- `repkey_timeout_handler()` (L190) calls `osal_del_timer(&repkey_timeout_timer)` at **L192** — a timer callback deleting its own timer. `del_timer_sync()` there waits for the handler to finish from inside the handler: instant deadlock.
- `hiir_interrupt()` (L196) calls `osal_del_timer()` at **L216** and **L243** — hard IRQ context, where `del_timer_sync()` must not be called at all.

So flipping `osal_del_timer()` to the sync variant breaks the IR driver in two separate ways.

## Suggested split

1. **`osal_timer_destory()` → `del_timer_sync()` before `kfree()`.** This is where the memory is actually freed, so it is where the hazard lives, and it is safe to make synchronous: destroying a timer from inside its own callback is *already* a use-after-free today, so no correct caller can be doing it. Worth confirming callers are in process context — `rtc_exit()` is.
2. **Leave `osal_del_timer()` as `del_timer()`.** Callers legitimately use it from IRQ context and from the callback itself to mean "stop re-arming".
3. Optionally add **`osal_del_timer_sync()`** for callers that want synchronous teardown and know they are in process context.

Callers should also keep a "stopping" flag checked before re-arming, since a self-re-arming callback can otherwise re-queue after the delete but before the free. That is the pattern #209 ended up using.

## Blast radius

Beyond the two source drivers above, the API is linked by prebuilt vendor objects across `vi`, `vpss`, `chnl`, `pm`, `rtc`, `ir`, `vdec` and `dis` on gk7205v200, hi3516ev200, hi3516cv300, hi3516cv500, hi3516cv6xx and hi3519v101 — core video path, not a corner. I could not audit those for the self-re-arming pattern, but fixing `osal_timer_destory()` would harden them without touching the blobs, which is the leverage this repo has.

## Not verified

I have not reproduced a crash. The window is a fraction of a timer period and a negative result would not mean much; the argument here is from the source. Nor have I audited the prebuilt objects' callbacks.

widgetii:
Closes #210.

## The bug

`osal_del_timer()` and `osal_timer_destory()` are both `del_timer()`, in the `COMPAT_TIMER_SETUP` branch and the pre-4.15 one alike, and destroy `kfree()`s the timer on top. `del_timer()` returns as soon as the timer is off the list — it does not wait for a callback already running on another CPU. A teardown that lands while the callback runs therefore frees the timer underneath it, and a callback that re-arms itself then schedules freed memory.

`kernel/rtc/hi3519v101/hi_rtc.c` is exactly that shape: `temperature_detection()` is the callback (L999) and ends by re-arming itself (L803), while `rtc_exit()` deletes and destroys (L1062). It ships as a blob, so it cannot be given a stop flag from here.

## The change

Make the destroy path wait. Two constraints shaped it:

**`del_timer_sync()` cannot be used from interrupt context** — it spins for the handler and deadlocks when that handler is on this CPU. Keep `del_timer()` there and name it in the log, because destroying a timer from an interrupt cannot be made safe in the first place, and a hang would be worse than what it replaces.

**`del_timer_sync()` alone is not enough against a self-re-arming callback** — it waits for the callback that has already put the timer back on the list. Repeat until a pass finds nothing pending, bounded so a callback re-arming with no delay cannot spin forever.

**`osal_del_timer()` is deliberately left alone.** This is the part worth checking me on. Callers use it from inside the timer's own callback *and* from hard IRQ context to mean "stop re-arming":

| driver | in the callback | in the IRQ handler |
|---|---|---|
| `ir/hi3516cv300/hiir.c` | L193 | L224, L251 |
| `ir/hi3516cv500/hiir.c` | L207 | L223, L264 |
| `ir/hi3519v101/hiir.c` | L192 | L216, L243 |

Making that one synchronous — which was the first suggestion on #210 — would deadlock all three. `osal_del_timer_sync()` is added instead, for callers in process context that need the callback finished before tearing down what it touches.

## Who this fixes

The two IR drivers built from source call `osal_timer_destory()` at exit and have callbacks that do not re-arm, so the destroy change covers them with no driver-side change.

The blob users get it for free, which is the point — they cannot be patched, only the code under them can: `vi`, `vpss`, `chnl`, `pm`, `rtc`, `vdec`, `dis` across gk7205v200, hi3516ev200, hi3516cv300, hi3516cv500, hi3516cv6xx and hi3519v101.

## Verification

Built the way the firmware builds it — buildroot, `gk7205v200_lite`, with `HISILICON_OPENSDK_OVERRIDE_SRCDIR` pointed at this tree — and run on a lab gk7205v200. 104188 bytes against the shipped 103400, vermagic identical, and it boots the full 36-module stack with video serving.

| exercise | result |
|---|---|
| boot with the patched osal | 36 modules, `open_osal 63022` with 60 users, JPEG off the camera |
| 10 × `load_goke -a` (full teardown + reload — every osal timer user destroyed and re-created) | 36 modules back every cycle, no rmmod refusals |
| 8 × SIGHUP (pipeline rebuilds under majestic) | clean |
| final teardown + reload, then a capture | clean |

Zero oops, zero warnings across all of it. **No `osal_timer_stop` lines at all**, which is the informative part: on this part nothing destroys a timer from interrupt context, and nothing re-armed past the retry bound. Both new log lines stayed silent.

Camera restored to the shipped module afterwards and re-verified healthy.

One thing worth recording for anyone else testing kernel changes here: a module built *outside* buildroot is not loadable on the shipped firmware. My first attempt oopsed in `media_bus_init` during device registration — that was the build environment, not the change; a pristine build of the same tree was already 106824 bytes against the shipped 103400. Matching vermagic is necessary but not sufficient without `CONFIG_MODVERSIONS`. The buildroot override above is the way to do it.

Two new log lines make the residual cases visible rather than silent:

```
osal_timer_stop - called from interrupt, cannot wait for the timer callback
osal_timer_stop - timer keeps re-arming itself, gave up after 5 tries
```

If either shows up on another part, it identifies a caller worth looking at.

### DIFF
--- a/kernel/osal/include/osal.h
+++ b/kernel/osal/include/osal.h
@@ -534,6 +534,8 @@ extern int osal_hrtimer_destory(osal_hrtimer_t *phrtimer);
 extern int osal_timer_init(osal_timer_t *timer);
 extern int osal_set_timer(osal_timer_t *timer, unsigned long interval); // ms
 extern int osal_del_timer(osal_timer_t *timer);
+/* Waits for a callback already running; process context only. */
+extern int osal_del_timer_sync(osal_timer_t *timer);
 extern int osal_timer_destory(osal_timer_t *timer);
 
 extern unsigned long osal_msleep(unsigned int msecs);
--- a/kernel/osal/linux/kernel/osal_timer.c
+++ b/kernel/osal/linux/kernel/osal_timer.c
@@ -27,6 +27,50 @@ int osal_hrtimer_destory(osal_hrtimer_t *phrtimer)
 }
 
 /*
+ * del_timer() returns as soon as the timer is off the list; it does not wait
+ * for a callback already running on another CPU. Deleting and then freeing --
+ * which osal_timer_destory() does -- therefore leaves that callback writing
+ * into memory that has just gone away, and a callback that re-arms itself goes
+ * on to schedule a freed timer. OpenIPC/openhisilicon#210.
+ *
+ * del_timer_sync() waits, but it cannot be used from interrupt context: it
+ * spins for the handler, which deadlocks when the handler is on this CPU.
+ * Destroying a timer from an interrupt cannot be made safe at all -- there is
+ * no way to know the handler is not running -- so keep the old behaviour there
+ * and name it, rather than trading silent corruption for a hang.
+ *
+ * del_timer_sync() on its own is still not enough for a callback that re-arms
+ * itself: it waits for that callback, which has already put the timer back on
+ * the list by the time it returns. The rtc temperature poll does exactly this
+ * and ships as a blob, so it cannot be given a stop flag. Repeat until a pass
+ * finds nothing pending -- at that point no handler is running and nothing is
+ * queued -- and bound it, because a callback re-arming with no delay would
+ * otherwise spin here forever.
+ */
+#define OSAL_TIMER_STOP_TRIES 5
+
+static void osal_timer_stop(struct timer_list *t)
+{
+	int i;
+
+	if (in_interrupt()) {
+		osal_printk(
+			"%s - called from interrupt, cannot wait for the timer callback\n",
+			__FUNCTION__);
+		del_timer(t);
+		return;
+	}
+
+	for (i = 0; i < OSAL_TIMER_STOP_TRIES; i++) {
+		if (!del_timer_sync(t))
+			return;
+	}
+
+	osal_printk("%s - timer keeps re-arming itself, gave up after %d tries\n",
+		    __FUNCTION__, OSAL_TIMER_STOP_TRIES);
+}
+
+/*
  * On kernels 4.15+, timer_setup() replaces init_timer() and the callback
  * signature changes from void(*)(unsigned long) to void(*)(struct timer_list*).
  *
@@ -99,10 +143,28 @@ int osal_del_timer(osal_timer_t *timer)
 }
 EXPORT_SYMBOL(osal_del_timer);
 
+/* Callers legitimately use osal_del_timer() from an interrupt, and from inside
+ * the timer's own callback, to mean "stop re-arming"; neither can wait. This is
+ * the variant for callers in process context that need the callback finished
+ * before they tear down what it touches. */
+int osal_del_timer_sync(osal_timer_t *timer)
+{
+	struct osal_timer_compat *ot = NULL;
+	if ((timer == NULL) || (timer->timer == NULL) ||
+	    (timer->function == NULL)) {
+		osal_printk("%s - parameter invalid!\n", __FUNCTION__);
+		return -1;
+	}
+	ot = timer->timer;
+	osal_timer_stop(&ot->tl);
+	return 0;
+}
+EXPORT_SYMBOL(osal_del_timer_sync);
+
 int osal_timer_destory(osal_timer_t *timer)
 {
 	struct osal_timer_compat *ot = timer->timer;
-	del_timer(&ot->tl);
+	osal_timer_stop(&ot->tl);
 	kfree(ot);
 	timer->timer = NULL;
 	return 0;
@@ -160,10 +222,25 @@ int osal_del_timer(osal_timer_t *timer)
 }
 EXPORT_SYMBOL(osal_del_timer);
 
+/* See the COMPAT_TIMER_SETUP branch above. */
+int osal_del_timer_sync(osal_timer_t *timer)
+{
+	struct timer_list *t = NULL;
+	if ((timer == NULL) || (timer->timer == NULL) ||
+	    (timer->function == NULL)) {
+		osal_printk("%s - parameter invalid!\n", __FUNCTION__);
+		return -1;
+	}
+	t = timer->timer;
+	osal_timer_stop(t);
+	return 0;
+}
+EXPORT_SYMBOL(osal_del_timer_sync);
+
 int osal_timer_destory(osal_timer_t *timer)
 {
 	struct timer_list *t = timer->timer;
-	del_timer(t);
+	osal_timer_stop(t);
 	kfree(t);
 	timer->timer = NULL;
 	return 0;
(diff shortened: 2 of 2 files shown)


widgetii:
Runtime-verified now — I had said I couldn't, and that was wrong; the module just needed building the way the firmware builds it (buildroot with `HISILICON_OPENSDK_OVERRIDE_SRCDIR` at this tree) rather than standalone against a kernel tree.

On a lab gk7205v200: boots the full 36-module stack, then 10 × `load_goke -a` full teardown/reload cycles, 8 × SIGHUP pipeline rebuilds, and a final teardown — 36 modules back every time, zero oops, zero warnings, video serving throughout. Camera restored to the shipped module and re-verified.

Notably **no `osal_timer_stop` lines at all**: on this part nothing destroys a timer from interrupt context and nothing re-armed past the retry bound, so both defensive paths stayed silent. PR body and commit message updated with the details.

widgetii:
Both findings point at a real hole, and I have fixed it in 3fda72a — but the mechanism in the first one is not what is happening, so it is worth writing down.

## Finding 1: `del_timer_sync()` returning 0 on a self-re-arm

The claim is that `del_timer_sync()` "can occur while a callback is running and then re-arms the timer before `del_timer_sync()` returns", so the loop exits with the timer pending. That is not how it behaves. From 4.9 `kernel/time/timer.c`:

```c
int try_to_del_timer_sync(struct timer_list *timer)
{
	int ret = -1;
	base = lock_timer_base(timer, &flags);
	if (base->running_timer != timer) {
		ret = detach_if_pending(timer, base, true);
	}
	...
}

int del_timer_sync(struct timer_list *timer)
{
	for (;;) {
		int ret = try_to_del_timer_sync(timer);
		if (ret >= 0)
			return ret;
		cpu_relax();
	}
}
```

While the callback runs, `base->running_timer == timer`, so `try_to_del_timer_sync()` returns **-1** and `del_timer_sync()` spins. `base->running_timer` is cleared only after the callback returns. So for a callback that re-arms itself, the sequence is: spin while it runs → callback re-arms and returns → `running_timer` cleared → the timer is now pending → `detach_if_pending()` removes it and returns **1**.

`del_timer_sync()` returns 1, not 0, and the loop takes another pass, which then returns 0 with nothing pending and nothing running. That is exactly the case the loop was written for, and it is why it is bounded.

Where a timer really can come back after we return is a *different* context arming it — `hi_rtc`'s temperature interrupt does that alongside the callback. No amount of looping in the OSAL layer fixes that one; the caller has to stop the source first, which is what acodec does with a stopping flag in #209 and what I noted in the comment.

## What I did take, from both findings

The genuinely broken part, and it is the same defect underneath both: when `osal_timer_stop()` **could not** establish that the timer is off the list with no callback running, `osal_timer_destory()` went on and `kfree()`d anyway. That applies to the `in_interrupt()` fallback (finding 2) and to exhausting the retry bound (the tail of finding 1).

So `osal_timer_stop()` now returns a result, and destroy refuses to free on failure:

```c
if (osal_timer_stop(&ot->tl)) {
	osal_printk("%s - timer still live, leaked rather than freed\n", __FUNCTION__);
	return -1;
}
kfree(ot);
```

Leaking the wrapper is the lesser fault — a hundred bytes lost on a driver that is already misbehaving, against the use-after-free this set exists to prevent. `osal_del_timer_sync()` returns the same result so callers can tell.

I kept the `in_interrupt()` fallback rather than making it synchronous: `del_timer_sync()` from interrupt context spins for a handler that may be on this CPU, and 4.9 itself `WARN_ON(in_irq() && !TIMER_IRQSAFE)`s there. A hang would be worse than what it replaces. With this change that path no longer frees, so it can no longer be the use-after-free finding 2 describes — it degrades to a leak and a log line.

## Re-verified

Rebuilt through buildroot and re-run on the lab gk7205v200: boot to the full 36-module stack, 10 × `load_goke -a`, 8 × SIGHUP, capture after. Zero oops, zero warnings, and no `osal_timer_stop` or leak lines — the new refusal path stays unreached on this part, which is the result to want. Camera restored to the shipped module and re-verified.

PR #214, opened 2026-08-27
cipher: build open_cipher.ko for hi3516ev200

widgetii:
hi3516ev200 (and so hi3516ev300) is the one HiSilicon family whose OpenIPC image ships no cipher module at all. hi3516av300 and hi3516cv300 both package one and simply never load it; here there is nothing to load.

Two commits, because the first is a bug in its own right.

### `osal: give the hil_ MMZ names to source builds, not just to blobs`

`mmz/media-mem.c` already exports every `mmb_` function under both spellings via `EXPORT_COMPAT_MMZ` — `mmz_` for this tree, `hil_` for the vendor drivers that link against it. Two gaps in that:

- `mmz_mmb_getby_phys_2` was the one `mmb_` function left on a plain `EXPORT_SYMBOL`, so `hil_mmb_getby_phys_2` was never aliased. Every sibling above and below it uses the compat macro; this looks like an oversight rather than a decision. The symptom is a vendor cipher driver refusing to load with `Unknown symbol hil_mmb_getby_phys_2`, which is how it was found.
- `osal_mmz.h` carries no `hil_` names at all. That was enough while the only consumers were prebuilt `.o` blobs — they reference symbols and bring no headers. A driver built from vendor **source** needs the names at compile time too: its `drv_osal_lib.h` opens with `#include "osal_mmz.h"` and then a bare `hil_mmb_t`.

The header half is aliases only — a typedef and nine defines onto the `mmz_` names, nothing that allocates, frees or behaves differently. Chips whose osal comes from `kernel/include/<chip>/osal_mmz.h` already have these names natively and never reach this file, because the two share a basename and the include path picks one.

### `cipher: build open_cipher.ko for hi3516ev200`

The tree is the SDK's own (Hi3516EV200_SDK_V1.0.1.2), `include/` and `src/` only, without its `sample/`, `buildin/` and `patch/` directories.

Reusing `cipher/hi3516cv500/src` with `-DCHIP_TYPE_hi3516ev200` was tried first and is worse than it looks: cv500's is a later SDK vintage, the two differ by some 30k lines, and it carries no `drv_osal_hi3516ev200.h`. Borrowing cv500's headers is a trap of its own — `include/hi3516cv500/hi_types.h` shadows this tree's, which is where the `HI_HANDLE_*` macros live, and its `hi_type.h` declares `HI_PHYS_ADDR_T` unsigned where ev200 has it signed. The tree here is self-contained instead, `hi_type.h` included.

Built through the generic branch of `kernel/Kbuild` via a new `kernel/cipher/Kbuild`, guarded on `CHIPARCH` so only hi3516ev200 picks it up.

## Testing

On an hi3516ev300 (kernel 4.9.37), against the mbedTLS 2.25 the rootfs ships:

- module loads (`Load hi_cipher.ko success`), `/dev/cipher` appears
- a userspace harness drives it through **600 differential AES-CTR cases** — every length from 1 byte to 2048, byte-identical to mbedTLS, with nothing written past the requested length
- known-answer checks pass: NIST SP 800-38A F.5.1, and a vector across a 32-bit counter wrap (an engine that carries only the low 32 bits agrees with software for every IV that is not near a wrap, so a single ordinary vector would not catch it)
- at 1100-byte buffers, an SRTP-sized workload: **86.0 µs of CPU in software against 41.9 through the engine**, a 51% cut. More is saved in CPU than in wall clock, because the driver sleeps on its completion interrupt and gives the core back while the block works.

Build-checked that `CHIPARCH=gk7205v200` still produces no cipher module, and that hi3516cv500 — which has its own kbuild and its own cipher — is untouched.

## Not included

Not built for gk7205v200, which shares the generic branch: same generation on paper, but Goke ships its cipher as a differently-named module against a different userspace, and none of the above has been measured on one.

### DIFF
--- a/kernel/Kbuild
+++ b/kernel/Kbuild
@@ -61,6 +61,9 @@ include $(src)/sensor_i2c/Kbuild
 include $(src)/sensor_spi/Kbuild
 include $(src)/mipi_rx/Kbuild
 include $(src)/openipc_frame_ts/Kbuild
+# Guarded on CHIPARCH inside; only hi3516ev200 builds a cipher module here,
+# and hi3516cv500 builds its own from hi3516cv500.kbuild.
+include $(src)/cipher/Kbuild
 #+user
 
 ifneq ($(filter $(CHIPARCH),hi3516ev200 gk7205v200),)
--- a/kernel/cipher/Kbuild
+++ b/kernel/cipher/Kbuild
@@ -1,0 +1,91 @@
+#
+# kernel/cipher — the Cipher (crypto) engine, for chips built by the generic
+# branch of kernel/Kbuild rather than by their own <chip>.kbuild.
+#
+# hi3516cv500 builds this module from its own kbuild and does not come through
+# here. This file exists for hi3516ev200 (and so hi3516ev300), whose firmware
+# has never packaged a cipher module at all: the vendor prebuilt from the
+# Hi3516EV200 SDK cannot load against an OpenIPC rootfs, and there was no
+# source build to fall back on.
+#
+# ITS OWN TREE, like every other chip under kernel/cipher/. Reusing cv500's
+# source with -DCHIP_TYPE_hi3516ev200 was tried first and is a worse idea than
+# it looks: cv500's tree is a later SDK vintage (the 2019 rewrite), the two
+# differ by some 30k lines, and it carries no drv_osal_hi3516ev200.h at all.
+# The tree here is the one from Hi3516EV200_SDK_V1.0.1.2 that was built and
+# checked on an hi3516ev300 -- 600 differential AES-CTR cases byte-identical
+# to mbedTLS -- so it is the version known to drive this silicon correctly.
+# Only include/ and src/ are vendored; the SDK's sample/, buildin/ and patch/
+# trees are not built here.
+#
+# WHAT IT IS FOR. AES-128-CTR at SRTP packet sizes, where a userspace streamer
+# would otherwise do it on the CPU. Measured on an hi3516ev300 against the
+# mbedTLS 2.25 the rootfs ships, 1100-byte buffers: 87.6 us of CPU in software
+# against 42.0 through the engine, a 52% cut. More is saved in CPU than in
+# wall clock, because the driver sleeps on its completion interrupt and hands
+# the core back while the block works -- which is the scarce resource on these
+# parts. The engine also carries a TRNG and RSA that nothing uses yet.
+#
+# Not built for gk7205v200, which shares this branch: it is the same
+# generation on paper, but Goke ships its cipher as a differently-named module
+# against a different userspace, and nothing here has been measured on one.
+#
+
+ifeq ($(CHIPARCH),hi3516ev200)
+
+EV200_CIPHER = cipher/hi3516ev200/src
+
+$(PREFIX)cipher-objs := \
+	$(EV200_CIPHER)/drv/cipher_initdevice.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_symc_v100.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_symc_v200.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_hash_v100.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_hash_v200.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_ifep_rsa_v100.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_trng_v100.o \
+	$(EV200_CIPHER)/drv/cipher_v1.0/drivers/core/drv_trng_v200.o \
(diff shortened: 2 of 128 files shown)


widgetii:
Pushed two commits: the CI fix, and one of the five review findings.

## CI: `hi3516ev300_neo` (kernel 7.0)

My bug, and the split between rows is the tell — `hi3516ev200_lite` passed and only the 7.0 row failed, because the two take different halves of `EXPORT_COMPAT_MMZ`. On 6.4+ it expands to `typeof(mmz_x) hil_x __attribute__((alias("mmz_x")))`, which writes `hil_x` as a real identifier; the `#define hil_x mmz_x` I added rewrote that token and redefined `mmz_x` on top of itself. Below 6.4 the other branch builds its ksymtab entries out of strings and never spells `hil_x` as an identifier, so the macro was invisible there.

Now `extern typeof(...)` declarations, which work on both branches. Reproduced the exact error in a 20-line standalone file before and after, and rebuilt for 4.9 to confirm the passing row still passes.

One consequence worth flagging for anyone testing by hand: the cipher module now links against `hil_mmb_getby_phys_2`, the name the vendor source actually writes, where the `#define` quietly rerouted it to `mmz_mmb_getby_phys_2`. That name was already exported, so an unpaired cipher module used to load against an unpatched osal and now does not. The two ship from one build so it only affects ad-hoc swaps — but it cost me a confusing "unknown symbol" before I spotted it.

## Review findings

Thanks — all five are in the vendored HiSilicon SDK source rather than in code this PR authors, so I went through them on whether they are reachable and whether patching an import is the right response.

**Fixed: #2, interrupted wait reports success.** This one is real and it matters. `crypto_queue_wait_timeout` is `wait_event_interruptible_timeout` — the vendor's own comment two lines above its definition says "interrupted by a signal" — so `-ERESTARTSYS` is reachable whenever a signal reaches the thread inside the encrypt ioctl. The caller is then told the operation succeeded and copies out a buffer the engine may not have finished writing. Nothing downstream can catch that: ciphertext is meant to look like noise, so a partially-written buffer passes every check a caller could make and is discarded at the far end as ordinary packet loss. A streamer that reloads on SIGHUP is exactly the shape of program that would hit it.

Also **not new** — `hi3516cv500` has the identical line and ships it today on hi3516av300 — so both trees are fixed in that commit. If you would rather keep this PR to the ev200 import, drop it and I will send the cv500 half separately.

**Not fixed: #1, #3, #4, #5.** Reasoning rather than dismissal:

- **#4 and #5 (RSA)** are in `cryp_rsa.c`. Nothing in this build path touches RSA, and I have no way to exercise a fix on hardware.
- **#1 (failed mapping poisons retries)** is init-failure recovery, and **#3 (timeout leaves DMA active)** is timeout recovery. Both look accurate on reading. Neither is on the path this module is being added for, and for #3 specifically the consumer I have degrades safely: on a driver error it abandons the engine for that stream, and the DMA target is a static buffer that is never read again afterwards.

The general reason for stopping there: the same `-ERESTARTSYS` pattern also appears in the hash, `symc_v100` and IFEP-RSA waits in both trees, and mechanically rewriting vendor error handling that nothing here exercises is how a small fix becomes an unreviewable one. Happy to take any of them in a follow-up if you would rather have them in.

## Testing after the changes

Rebuilt both trees (ev200 against 4.9.37, cv500 against its own kernel). On an hi3516ev300, with the patched osal and cipher loaded together: 600 differential AES-CTR cases byte-identical to mbedTLS across every length from 1 to 2048, nothing written past the requested length, and 86.0us of CPU per 1100-byte buffer in software against 41.5 through the engine. Camera restored to its shipped modules afterwards.

widgetii:
CI is green on all sixteen rows.

The `hi3516ev300_neo` failure was the vendored tree's age, not the vendoring. `Hi3516EV200_SDK_V1.0.1.2` is 2019, and predates three kernel API changes that only the 7.0 row reaches:

* `platform_driver::remove` returns `void` from 6.11
* `proc_create()` takes a `struct proc_ops`, not a `struct file_operations`, from 5.6
* `of_dma_configure()` grew a third argument in 4.18

`bad4243` patched the cv500 cipher tree for exactly these three; `c3fd689` is that patch applied to ev200, taking the return type and statement from the `compat_platform_remove_ret` / `compat_platform_remove_return` macros the rest of the tree already uses. Only the `/proc/driver/hi_cipher` entry changes shape — `dev_cipher_fops` is a real character device and keeps `file_operations`.

All three are conditional, so the 4.9 build is textually unchanged after preprocessing; I rebuilt against the `hi3516ev300` 4.9 tree before pushing, and `hi3516ev200_lite` confirms it.

Worth saying what *isn't* here: I grepped the whole vendored tree for the rest of the 5.0–7.0 removals — `access_ok`, `set_fs`/`get_fs`, `DEFINE_SEMAPHORE`, `strlcpy`, `get_random_int`, `PDE_DATA`, `do_gettimeofday`, `mmap_sem`, `init_timer`, `dma_alloc_from_contiguous`, `register_sysctl_table`, `kmap_atomic`, `pte_offset_map`, `list_for_each` — and the only other hit was `ioremap_nocache`, which `kernel/compat/kernel_compat.h` already shims for the whole tree. The driver cores under `cipher_v1.0/drivers/` are register poking and touch no kernel API that has moved.

PR #223, opened 2026-09-12
wdt: make the watchdog reset the board when the thing it guards dies

### LINKED ISSUE #1803: Watchdog not working on hi3516ev300 and gk7205v300
Watchdog not working on hi3516ev300 (hi3516ev200 family), imx335, 16 MB nor flash, 2.5.04.22-ultimate and gk7205v300 (gk7205v200 family), imx335, 16 MB nor flash, 2.5.04.26-ultimate. When majestic crashes, the camera does not reboot automatically. I added the `nodeamon=1` parameter to the `modprobe open_wdt` command in `load_hisilicon` and `load_goke`.

https://github.com/OpenIPC/firmware/blob/b88ff6b39d5c7bf0ac7105ab563df042e4618912/general/package/hisilicon-osdrv-hi3516ev200/files/script/load_hisilicon#L199

https://github.com/OpenIPC/firmware/blob/b88ff6b39d5c7bf0ac7105ab563df042e4618912/general/package/goke-osdrv-gk7205v200/files/script/load_goke#L195

This made the camera reboot automatically when majestic crashes.

widgetii:
Fixes OpenIPC/firmware#1803.

The watchdog driver did not honour the Linux watchdog API in six ways, and the
driver exists ten times in this tree, so each one had to be found more than
once. This fixes all six and retires the duplication where the copies were the
same driver.

## The defects

Measured on lab boards, kernel 4.9.37 and 3.18.20.

**1. The margin meant twice itself.** `set_timeout()` loaded a whole margin into
a counter the SP805 runs twice — reaching zero raises the interrupt and reloads,
and only a second zero with that interrupt still pending drives the reset. A
margin of 30 s reset the board at 58 s. Loading half a margin is what
`drivers/watchdog/sp805_wdt.c` does.

**2. A ping did nothing for half the cycle.** `feed()` read `WDT_RIS` first and
returned without touching anything when no interrupt was pending, so the
deadline stayed where the previous accepted ping had put it. The honest cadence
of one ping per half-margin was therefore worth exactly one accepted ping per
cycle: miss it and a process that had been feeding the device all along got the
board reset under it. The cv500 and V5 copies were worse — even when the
interrupt *was* pending they only cleared it, never reloading.

**3. An unexpected close handed the dog back to the driver.** This is #1803.
The driver runs a kernel thread that feeds the watchdog whenever `dog_state` is
`SELFCLR`; `open()` set `EXTCLR` so the thread stood down, but `release()` set
`SELFCLR` again — unconditionally. So the one event the device exists for, the
guarded process dying and the kernel closing its descriptor, handed the watchdog
straight back to the driver, which then fed it for the rest of the board's life.
Nothing ever reset. The only workaround available was `nodeamon=1`: no feeder at
all, and therefore no cover for the module-load-to-first-open window either.

`WDIOF_MAGICCLOSE` has been advertised throughout, and it means a close without
a preceding `"V"` leaves the timer running. The feeder now covers the pre-open
window and nothing else — `open()` latches `EXTCLR` one-way and `release()` no
longer undoes it — and a write handler makes the magic close reachable at all.

| board | before | after |
| --- | --- | --- |
| hi3516ev300 | majestic SIGKILLed at uptime 40195, still up at 40400 | reset at 27–36 s |
| hi3516av300 | SIGKILLed at 215043, still up at 215249 | reset at 27–36 s |
| gk7205v200 | probe died at uptime 28050, still up 164 s later | reset at 27–36 s |

**4. `open()` did not start the timer.** A magic close, or `WDIOC_SETOPTIONS`
with `WDIOS_DISABLECARD`, clears `WDT_CTRL`, and neither the ping in `open()`
nor `WDIOC_SETTIMEOUT` puts it back. From the first deliberate stop the device
was inert while looking entirely healthy: the open succeeded, the margin was
accepted and read back, the pings were accepted, and nothing would ever have
reset the board. majestic did not notice because it sends `WDIOS_ENABLECARD`
itself right after opening; a client that only does what the API documents
(busybox watchdog, systemd) got nothing.

**5. `rmmod` could wedge the board permanently.** `watchdog_exit()` stopped the
feeder before unregistering the device. `dog_exit()` sets `DOG_EXIT` and then
blocks in `kthread_stop()` waiting for the thread to notice — and the thread's
loop condition is `dog_state`, which `open()` sets straight back to `EXTCLR`.
One open racing the unload left the thread looping for ever:

```
rmmod   D  kthread_stop
           osal_kthread_destory [open_osal]
           watchdog_exit [open_wdt]
           wdg_remove [open_wdt]
           platform_drv_remove
           SyS_delete_module
```

`module_mutex` held, the task unkillable, and the board did not survive a clean
reboot — it needed the power pulled. Hit on a lab gk7205v200 by nothing more
exotic than running a probe against `/dev/watchdog` while an `rmmod` was in
flight, and unrecoverable once lost: `release()` leaves the state at `SELFCLR`
and `open()` at `EXTCLR`, neither of which is `DOG_EXIT`. Unregistering first
closes the window. Verified on the same board — with a shell loop opening and
closing `/dev/watchdog` as fast as it can, `rmmod` now returns 0 in two seconds.

**6. hi3516cv300 had no watchdog at all.** `wdt/hi3516cv300/hi_wdt.c` contained
no module entry point and its kbuild linked no init object, so
`hi3516cv300_wdt.ko` loaded, did nothing, and the board came up with **no
`/dev/watchdog`**. majestic logged `Watchdog device (/dev/watchdog) not found`
and every cv300 ran unguarded. Confirmed on a lab hi3516cv300 before this
change.

Smaller things alongside: `WDIOC_GETTIMELEFT` (was `EPERM` everywhere);
`-EINVAL` for an out-of-range margin instead of clamping to one nobody asked
for; `SETTIMEOUT` reading back the margin in force; `-ENOTTY` for an unknown
ioctl; and a feeder paced on the margin in force rather than the module default,
so a caller that lowers the margin and hands the device back is not reset on a
cycle nobody configured.

## The consolidation

hi3516cv500 and hi3516cv300 now build `wdt/wdt.c` and `wdt/wdt_init.c`. What
actually differs comes from the kbuild:

| | |
| --- | --- |
| `WDT_BASE` | the register map |
| `WDT_OSAL_HI` | OSAL is `hi_osal.h` here, not `osal.h` |
| `WDT_OSAL_NO_KTHREAD_STOP_FLAG` | V3's `osal_kthread_destory()` takes the task alone and always stops it |
| `WDT_NO_OF` | no watchdog node in the device tree — register on load and map `WDT_BASE` directly |
| `WDT_SCTL_BASE` | the 3 MHz clock is gated in the system controller and needs ungating first |

`wdt_init.c` carries both binding styles and names the device-tree compatibles
outright rather than assembling one from `compat.h`, which only knows the two V4
chips; that retires `init/hi3516cv500/wtdg_init.c` too. `DOG_TIMER_MARGIN` moved
into `wdt/watchdog.h`, where `wdt_init.c` can see it — the parameter description
used to stringify the undefined token and print `(default=DOG_TIMER_MARGIN)`.

Net −837 lines.

## Verified

Four boards, both binding styles. Crash path is open, `SETTIMEOUT(30)`, die
without `"V"`; the board must reset one margin later and come back.

| board | binding | base | crash → reset | `GETTIMELEFT` | magic close | clean stop |
| --- | --- | --- | --- | --- | --- | --- |
| hi3516cv300 | `WDT_NO_OF` | `0x12080000` | 30 s (was: no device) | 299/300 | 1 | no reset |
| hi3516av300 | platform | `0x12051000` | 30 s (was: never) | 299/300 | 1 | no reset |
| hi3516ev300 | platform | `0x12030000` | 30 s (was: never) | 299/300 | 1 | no reset |
| gk7205v200 | platform | `0x12030000` | 30 s (was: never) | 299/300 | 1 | no reset |

Soaks with majestic feeding at its own cadence, past the 300 s margin: cv300
456 s, av300 491 s, ev300 1327 s, no reboot. On cv300 majestic now logs
`Detected Watchdog` / `Watchdog timeout set to 300 seconds` for the first time.

Builds clean for hi3516cv300, hi3516cv500, hi3516ev200 (both `output-ev200` and
`output-hi3516ev300`) and gk7205v200 (both `output-gk` and `output-gk3`).

## Not consolidated, and why

- **hi3516cv6xx, hi3519dv500** — a later vendor generation: an optional FIQ
  feeding path, an hrtimer, `ot_osal.h`, and a command-list ioctl dispatch
  rather than `unlocked_ioctl`. All six fixes are carried over to both in place
  instead. **There is no firmware tree on this machine for either, so those two
  files are reviewed but neither compiled nor run** — the one untested part of
  this branch, and the thing most worth a second pair of eyes.
- **hi3516cv100, hi3519v101** — their kbuild links a vendor `.o`; the source
  beside it is reference and reaches no board. Benefiting needs a blob→source
  switch first.
- **hi3516cv200, hi3516av100, hi3520dv200** — V2 has no OSAL at all (the
  `osal_v2_shim` is a kernel-API compat shim for blobs, not an OSAL), and the
  driver there is written against plain kernel APIs. Sharing `wdt.c` would mean
  writing an OSAL for the generation first. `WDT_SCTL_BASE` exists in `wdt.c`
  ready for them.

Also related: OpenIPC/firmware#127 asks for the watchdog module on
hi3516cv200/hi3518ev200. It is already `insmod`ed there (the `#nodeamon=1`
beside it in `load_hisilicon` is a comment, not a parameter) — but that `.ko` is
a prebuilt vendor blob in the firmware package, so none of this reaches those
boards until it is built from source.

## Worth a release note

cv300 goes from having no watchdog to having a working one. That is the fix, but
it does mean a cv300 whose majestic wedges will now reboot where it previously
sat there. Benign, and the point of the device, but it is a behaviour change in
the field.


widgetii:
All five findings were real and are fixed in 88528b4. Two of them were mine and my testing had masked them — every probe set a timeout after opening, which is exactly the step that hid the problem.

**2 and 3 — reopen ignores the configured timeout.** Correct, and the more serious of the five. `dog_stop()` disarms via `dog_set_timeout(0)`, which parks `load_val` at the counter maximum; the `dog_start()` I added to `open()` then only enabled the counter and the ping wrote that maximum back. So `WDIOC_GETTIMEOUT` kept reporting the margin the client had asked for while the hardware ran at roughly the whole range. Reproduced on all four lab boards before the fix:

```
run1: SETTIMEOUT(30) -> 30
run1: stopped with magic close
run2: GETTIMEOUT  -> 30      <- what the client is told
run2: GETTIMELEFT -> 1431    <- what the hardware is doing
```

`open()` now restores the heartbeat rather than pinging, which is what the `WDIOS_ENABLECARD` path already did for the same reason. After, on all four: `GETTIMEOUT -> 30`, `GETTIMELEFT -> 29`.

**1 — a failed load leaves the dog armed.** Correct. `dog_init()` arms before creating the feeder and returned failure without disarming, so `watchdog_init()` unmapped the registers and failed the load with the counter still running and nothing in the system able to feed it — a reset one margin later, every boot. `dog_stop()` on both feeder-creation failure paths. The note that hi3516cv300 newly reaches this code is right; it previously had no module entry point at all.

**4 and 5 — `expect_close` survives a faulting write.** Correct that the flag is left set when a later byte faults, so a close after a partly-failed write is taken as deliberate and stops the watchdog. Worth noting the watchdog core has the same shape (`watchdog_write()` returns `-EFAULT` mid-loop without clearing `_WDOG_ALLOW_RELEASE`), so this is a deliberate deviation rather than a straight bug fix — but here the failure mode is a board that quietly stops being guarded, which is worth not inheriting. Cleared on the error path.

All three changes are in `wdt.c` and in both later-generation copies.

On the alternative approaches in the summary: agreed that migrating to the Linux watchdog framework is the right long-term answer and the wrong thing to combine with a behavioural fix for these vendor kernels. Runtime platform data would also be a reasonable shape, but every target already builds a SoC-specific module, so the compile-time flags cost nothing here and keep the no-device-tree path simple.

One correction to the PR description while I am here: `hi3519dv500` does build in CI and passed on 63e6abe, so it is no longer untested. `hi3516cv6xx` is genuinely not built anywhere — it is not in the CI matrix and there is no firmware tree for it locally — so for that one file I have only a syntax-only compile against its own vendor OSAL and kernel headers, plus the fact that its sibling `hi3519dv500` builds from an identical patch. That remains the weakest point of the branch.

